PR #64, opened 2015-11-30
Adding VerticalSurroundings

### LINKED ISSUE #38: Add values to Surroundings for stacked multi-family units
I don't see a good way to indicate that a unit in a multi-family home is stacked above or below the other units in the building.  The //BuildingSummary/Site/Surroundings element has a choices for how many sides the unit is attached on, but doesn't indicate if it is above or below other units.

I wonder if an easy way to indicate the stacked configuration is to add three enumerations to this element for:
stacked below other units
stacked between other units
stacked on top of other units

What do you think?

nmerket:
Fixes #38 
What do you think of this @brandongallagher @GamalielL?
I still don't like the names, and am open to alternatives.

### DIFF
--- a/schemas/HPXMLDataTypes.xsd
+++ b/schemas/HPXMLDataTypes.xsd
@@ -309,6 +309,14 @@
 			<xs:enumeration value="attached on three sides"/>
 		</xs:restriction>
 	</xs:simpleType>
+	<xs:simpleType name="VerticalSurroundings">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="unit above"/>
+			<xs:enumeration value="unit below"/>
+			<xs:enumeration value="unit above and below"/>
+			<xs:enumeration value="no units above or below"/>
+		</xs:restriction>
+	</xs:simpleType>
 	<xs:simpleType name="ShieldingofHome">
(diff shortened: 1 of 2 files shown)


schemas/HPXMLDataTypes.xsd
```diff
@@ -309,6 +309,14 @@
 			<xs:enumeration value="attached on three sides"/>
 		</xs:restriction>
 	</xs:simpleType>
+	<xs:simpleType name="VerticalSurroundings">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="unit above"/>
+			<xs:enumeration value="unit below"/>
+			<xs:enumeration value="unit above and below"/>
+			<xs:enumeration value="no units above or below"/>
```
nmerket: Here are the added enumerations.

GamalielL: I don't know if we can really improve on the name given that we have
SiteSurroundings describing the same thing horizontally.  That was the time
to pick a better name.  This looks good to me.

brandongallagher: Looks good to me too.

-brandon

PR #68, opened 2015-12-01
adding static pressure measurement elements

### LINKED ISSUE #33: Add Supply and Return Duct Static Pressures
Submitted by APS. Justification to follow.

nmerket:
baseelements_airdistributioninfo

Fixes #33 
@wmurphy67 @ChrisCBaker

### DIFF
--- a/schemas/HPXMLDataTypes.xsd
+++ b/schemas/HPXMLDataTypes.xsd
@@ -1044,6 +1044,19 @@
 		<xs:restriction base="xs:double"/>
 	</xs:simpleType>
 	<!--HVAC System Below-->
+	<xs:simpleType name="AirHandlerStaticPressureMeasurementLocation">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="in ducts"/>
+			<xs:enumeration value="at equipment"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:simpleType name="StaticPressureSource">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="as measured"/>
+			<xs:enumeration value="per design report"/>
+			<xs:enumeration value="per OEM documentation"/>
+		</xs:restriction>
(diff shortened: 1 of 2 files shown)


schemas/HPXMLDataTypes.xsd
```diff
@@ -1044,6 +1044,19 @@
...
+			<xs:enumeration value="at equipment"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:simpleType name="StaticPressureSource">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="as measured"/>
+			<xs:enumeration value="per design report"/>
+			<xs:enumeration value="per OEM documentation"/>
+		</xs:restriction>
+	</xs:simpleType>
```
nmerket: Here are the enumerations

wmurphy67:
I would recommend naming the field "total external static pressure measurement".

Naming it just static pressure measurement does not describe what we are actually asking for in the data field.

Casey Murphy

wmurphy67@mac.com
240.375.2311 (m)

ChrisCBaker:
I agree with the heading of "Total External Static Pressure", but do want to be clear that both supply and return pressures should be recorded separately as that can show problems that may exists where a total TESP test may not.

nmerket:
Okay, I changed the name of the element. It looks like this now:

baseelements_airdistributioninfo

@ChrisCBaker What should an annotation say that could go in the schema on that element to clarify the intent to have both supply and return pressures? Should we be allowing an unbounded number of these measurements for a duct system or just two -- one for supply and one for return?

ChrisCBaker:
I think just two measurement per system works. One for supply, one for return. Could annotation be “Ret - ” and “Sup +”?

Chris

nmerket:
I think this should make things clearer.

baseelements_totalexternalstaticpressuremeasurement

ChrisCBaker:
Looks good.

PR #69, opened 2015-12-01
Adding geothermal desuperheater water heater type

### LINKED ISSUE #51: Add Geothermal desuperheater
Add this water heater type

nmerket:
Fixes #51
Should probably run this by everyone in case it's not this simple.
@RebHudson


nmerket:
Make a separate element instead of enumeration. Boolean.

PR #71, opened 2015-12-03
Wind System

### LINKED ISSUE #52: Add wind
Add wind power into the schema in a similar way as PV

nmerket:
Here's my straw man at a wind system.

baseelements_wind

#### Thoughts

 - Do we need `RatedCapacity`? Isn't that covered by the other AWEA elements?
 - I left off actual annual output because we're unlikely to ever get it and the point is to document the physical characteristics anyway.

Fixes #52 
@RebHudson

### DIFF
--- a/schemas/HPXMLDataTypes.xsd
+++ b/schemas/HPXMLDataTypes.xsd
@@ -2012,4 +2012,10 @@
 			<xs:minInclusive value="0"/>
 		</xs:restriction>
 	</xs:simpleType>
+	<xs:simpleType name="WindThirdPartyCertification">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="AWEA 9.1-2009"/>
+			<xs:enumeration value="other"/>
+		</xs:restriction>
+	</xs:simpleType>
 </xs:schema>
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -1464,6 +1464,52 @@
 					</xs:sequence>
 				</xs:complexType>
 			</xs:element>
+			<xs:element minOccurs="0" name="Wind">
+				<xs:complexType>
+					<xs:sequence>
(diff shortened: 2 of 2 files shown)


schemas/HPXMLDataTypes.xsd
```diff
@@ -2012,4 +2012,10 @@
 			<xs:minInclusive value="0"/>
 		</xs:restriction>
 	</xs:simpleType>
+	<xs:simpleType name="WindThirdPartyCertification">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="AWEA 9.1-2009"/>
+			<xs:enumeration value="other"/>
+		</xs:restriction>
```
nmerket: enumerations for 3rd party certs.

RebHudson:
I dont think you need both rated capacity and peak power. They seem to be the same measurement.

nmerket:
That's what I was thinking but wanted to check.

nmerket:
Review:
@juliecaracino

juliecaracino:
Spoke with Greg Matzat, senior advisor, large scale renewables at NYSERDA (formerly DOE), and he recommended the following changes:

1. RatedCapacity and AWEARatedPower are the same (looks like you deleted one)
2. Add Rotor Diameter
3. Add Hub Height
4. Add Annual Average Power Output

Also, recommended we look at DOE's 2014 Distributed Wind Market Report http://energy.gov/eere/wind/downloads/2014-distributed-wind-market-report for guidance.

nmerket:
@juliecaracino I added `RotorDiameter` and `HubHeight`, but left off Annual Average Power Output because it wasn't clearly defined and it can be easily calculated from `AWEARatedAnnualEnergy`. Is that okay?

juliecaracino:
@nmerket Yes, that is OK. thanks!

juliecaracino:
@nmerket 

RESO has 2 more elements - seller owned and third party owned...should we add these?

PR #73, opened 2015-12-17
Adding renewable ready characteristics

### LINKED ISSUE #55: Renewable ready characteristics
Add the following renewable ready characteristics: 
Pre-wired to install system
Evaluated for solar availability
Electrical panel can handle a system

nmerket:
Updated: Click on the image to see it bigger.

baseelements_renewableready

fixes #55 
@RebHudson


nmerket:
This makes `PVSystem` optional in `Systems/Photovoltaics` and adds all these elements to cover what's in the checklist. There's a lot of pretty subjective stuff, I'll be interested to hear others' opinions on this.

nmerket:
[Here's the form](https://www.energystar.gov/ia/partners/bldrs_lenders_raters/rerh/docs/RERH_Guidance_PV_Checklist.pdf?d219-aaee) it provides data elements for. 

Review:
@RebHudson

juliecaracino:
@nmerket  

Are the elements under PV system still there with these proposed changes. Many of the new elements seem redundant. Can we add the elements on the form not already in HPXML under PV systems and then do a mapping to create the form? Rather than reproduce the form in HPXML?

nmerket:
Yes, it's all still there. 

hpxml_photovoltaics

It seems to me the only potentially redundant thing is the proposed array information and that's different than the regular `PVSystem` in that it is for the actually installed one.

RebHudson:
DOE has a slightly shortened version for Zero Energy Ready Homes - and it removes all of section 4 homeowner education. I'm open to removing that section unless that would be useful to any programs that want to monitor that component.

RebHudson:
I'm going to make a list of the items I find useful and the others I think could be useful and hopefully cut down on some of this

nmerket:
@RebHudson I guess my thought is why not have the data points there if people are collecting them in a form? If they don't include them in a particular document, that's fine too.

RebHudson:
@nmerket I'm fine with that - I see a couple items missing, but I'm fine not removing any. It seemed like there might be interest in fewer items.

RebHudson:
I don't see 3.3 and 3.4 
3.3 Install a 1” metal conduit from designated inverter location to electrical service panel (cap and label both ends). 
3.4 Install and label a 70-amp dual pole circuit breaker in the electrical service panel for use by the PV system (label the service panel).

RebHudson:
I would use 1.4, 1.5, 3.2, 3.3, 3.4 (Possibly 3.1). The DOE version handles 1.4/1.5 differently so it also has a question to assess viability: "Location, based on zip code has at least 5 kWh/m2/day average daily solar radiation based on annual solar insolation using PVWatts online tool: http://gisatnrel.nrel.gov/PVWatts_Viewer/index.html" I could also use this value (either as a boolean or a numerical value from PVWatts) in conjunction with 1.4 (if the contractor did this calculation and not the one in 1.5)

nmerket:
@RebHudson Good catch. I added 3.3-3.5. I'm not really sure what to do with your comment about the differences w/ the DOE version. Do I need to change anything?

RebHudson:
I just noticed that there were two checklists in use and if the DOE checklist is used, they would have a different but overlapping question related to availability of solar. I don't know if adding that data point is valuable or not. I'm not sure if anyone who does Zero energy ready homes is on the working group and would have an opinion on that.

nmerket:
This seems like its no longer a need. I'm closing this PR. If you disagree, let me know and we can reopen it.

PR #76, opened 2016-02-01
adding year and month element to when the duct system was sealed

### LINKED ISSUE #45: Have ducts sealed associated with a specific year completed
Add the association with the year with the job task

nmerket:
Adding this element

baseelements_hvacdistributionimprovementinfo

It requires a valid in the form `YYYY-MM` (i.e. `2016-02`). Is that too granular @RebHudson?

fixes #45


RebHudson:
That is more granular than I need but if it makes sense to have the month, I am fine keeping the month

PR #93, opened 2016-06-13
Adding CO, smoke, and radon detectors

### LINKED ISSUE #83: Carbon Monoxide Detectors
I don't see that there is a way to indicate that a CO detector is present and/or how many CO detectors are present.  This seems like something we should add.  

I think this should be an integer field to specify how many there are.

It seems like it should go under HealthAndSafety.  It probably should go under the CombustionApplianceZone node, but since this is an array node, there could be multiple CAZ's, so it might be unclear which one a CO detector belongs under.  Maybe the CODetector could go directly under the HealthAndSafety node?  Or maybe there could be a general node for "Detectors" that could include CO, smoke, and Radon?

nmerket:
baseelements_healthandsafety

fixes #83

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -2398,6 +2398,32 @@
 					</xs:sequence>
 				</xs:complexType>
 			</xs:element>
+			<xs:element minOccurs="0" name="Detectors">
+				<xs:complexType>
+					<xs:sequence>
+						<xs:group ref="SystemInfo"/>
+						<xs:element minOccurs="0" name="Smoke"
+							type="IntegerGreaterThanOrEqualToZero">
+							<xs:annotation>
+								<xs:documentation>Number of smoke detectors</xs:documentation>
+							</xs:annotation>
+						</xs:element>
+						<xs:element minOccurs="0" name="CarbonMonoxide"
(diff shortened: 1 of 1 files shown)


nmerket:
Review @brandongallagher

nmerket:
@tglidden
@ejacobsohn

nmerket:
Need to:
1. Figure out use case
2. Figure out data elements and structure

nmerket:
From Kevin Powell:

>I believe the original question was asking about including CO monitors, and then followed up by asking if radon and smoke detectors should also be included and, if so, how would it be done such as by zone or what.  This may be more complicated than a simple “add by device” in that I think we should talk to Noel to see what implications exist if they need to be by zone (I’ll check into building code as I know that there is a separate criteria by level and existence of a bedroom irrespective of number of HVAC zones.  Additionally, we need to check if placement (height, location on wall, etc) needs to be factored into the equation and if HPXML can handle this addition lift.

We can add a reference to the zone the same as we do for other things. We can handle whatever level of complexity we need to in HPXML, but I just need to know what that is and I don't right now.

lirainer:
Note that there are devices that combine smoke and CO detection. Should those be supported?

PR #100, opened 2016-08-12
Implementation guide edits

nmerket:
Rendered guide can be found [here](http://hpxml-guide.readthedocs.io/en/docchanges/).


nmerket:
@juliecaracino I'm going to hold off on deleting sections until we talk again just to make sure I don't cause problems.

PR #104, opened 2017-07-24
Adding preconstruction event type

nmerket:
From David Heslam:

>It would be great if you ask the working group about adding another enumeration for the HES Assessment Type in HPXML called "preconstruction" with a description of "An assessment based on plans for a house under construction".

### DIFF
--- a/schemas/HPXMLDataTypes.xsd
+++ b/schemas/HPXMLDataTypes.xsd
@@ -529,6 +529,7 @@
 			<xs:enumeration value="construction-period testing/daily test out"/>
 			<xs:enumeration value="job completion testing/final inspection"/>
 			<xs:enumeration value="quality assurance/monitoring"/>
+			<xs:enumeration value="preconstruction"/>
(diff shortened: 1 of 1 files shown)


Tedkidd:
Should there be sub labels? 

When we are involved in new construction, we are there pre-drywall to discuss air sealing strategies/opportunities, post drywall for preliminary blower door diagnostics, post HVAC for  thermostat setup including static testing (usually a Carrier Infinity Communicating Stat), and at completion for final blower door and HVAC. 

There are data and observation that would be great to record at each point.

vtnate:
I like the concept of organizing and categorizing. Is this meaningfully different from the "approved workscope" value on line 528? Could that value be used instead and not lose critical information?

nmerket:
@Tedkidd @vtnate The problem with adding sub labels at this point is that it makes this a breaking change. If we want to tackle a reorganization of the event types we should do it as part of v3.0.

PR #108, opened 2017-08-30
Additional appliance fields

shorowit:


brandongallagher:
I don't have much to say about many of these fields, since I don't know what HERS requires, but I noticed the RatedAnnualkWh field that got added to the ClothesWasherInfoType.

I wonder if it would be a good idea to add RatedAnnualkWh to all appliances?  It could be a useful way to document a simple, standard measure of the appliance's efficiency.  It seems like most or all of the appliances we care about would have an energy label with this info on it.

If this seems like a good idea, then I suggest we move the RatedAnnualkWh field into the ApplianceTypeSummaryInfo element.

PR #109, opened 2017-08-30
Clarify dhw distribution fields

shorowit:


shorowit:
Not sure if we actually want RESNET-specific documentation in the schema. Perhaps it should just live in the ERI validator?

PR #113, opened 2017-11-17
Add Solar Absorptance & Emittance

shorowit:
Adds solar absorptance and emittance elements for non-foundation walls/roofs and rim joists. Needed for an ERI calculation. Note that this branch is based off of the `attics_refactor` branch.


shorowit:

PR #137, opened 2018-07-30
Adding FloorFurnace

### LINKED ISSUE #101: Add new HeatingSystemType/FloorFurnace/extension/Any enumeration
This type of system is common in some areas and currently causes problems with HES generation. OptiMIser designates floor furnaces using the HeatingSystemType/Other/Description enumeration.  This designation fails in the HPXML-HES translator.  If we had an a FloorFurnace enumeration, then the HPXML-HES translator could map it to the Wall Furnace option in the HES inputs.  Glenn Dickey has approved the use of WallFurnace as a workaround.  I'm currently instructing users to make that designation in the interface.  I could remap our HPXML translation, so that Floor Furnaces are specified as Wall Furnace in the xml, but I think it would be much better to have a Floor Furnace designation in HPXML.

nmerket:
This adds the `FloorFurnace` heating system type with the same subelements as `WallFurnace`.

baseelements_heatingsystemtype

Fixes #101

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -2763,16 +2763,8 @@ NOTE: This is for ALL measures that may create entry points for vermin. For exam
 					</xs:sequence>
 				</xs:complexType>
 			</xs:element>
-			<xs:element name="WallFurnace">
-				<xs:complexType>
-					<xs:sequence>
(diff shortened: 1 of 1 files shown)


GamalielL:
Doesn't adding the WallAndFloorFurnace Type and applying it to WallFurnace make this a breaking change?

nmerket:
No. The subelements of `WallFurnace` all stay the same.

PR #144, opened 2018-09-01
Clothes Washer IMEF/IWF

shorowit:
Adds clothes washer fields for IMEF/IWF, the new EnergyStar performance metrics as of 3/7/15.

image


shorowit:
Add notes for MEF/WF.

PR #145, opened 2018-09-01
Clothes Dryer CEF (and EF bugfix)

shorowit:
Adds the newer CEF metric for clothes dryers. Also fixes EF to be `EnergyFactor`, not `EfficiencyFactor`.

BaseElements_ClothesDryer

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -232,9 +232,14 @@
 							<xs:documentation>loads/week</xs:documentation>
 						</xs:annotation>
 					</xs:element>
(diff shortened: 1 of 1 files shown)


shorowit:
I started to second guess myself on whether EF stands for Energy Factor or Efficiency Factor. The ANSI/RESNET/ICC standard says "efficiency factor", and you can find a lot of references to that term in Google. 

But the DOE Standards rulemaking makes it clear that it's "energy factor". See https://www1.eere.energy.gov/buildings/appliance_standards/standards.aspx?productid=50&action=viewlive, click on "78 FR 49608", and search for "energy factor".

shorowit:
Add note for EF.

PR #159, opened 2019-01-18
Adds fuel type element for heat pumps

shorowit:
For example, heat pumps powered by natural gas exist. This change also makes it easier for software developers to loop through all HVACPlant system types and grab the fuel type.

image

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -571,9 +571,9 @@
 												<xs:element minOccurs="0" name="RoofColor"
 												type="WallAndRoofColor"/>
 												<xs:element minOccurs="0" name="SolarAbsorptance"
-													type="SolarAbsorptance"/>
+												type="SolarAbsorptance"/>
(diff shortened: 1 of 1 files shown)


schemas/BaseElements.xsd
```diff
@@ -571,9 +571,9 @@
 												<xs:element minOccurs="0" name="RoofColor"
 												type="WallAndRoofColor"/>
 												<xs:element minOccurs="0" name="SolarAbsorptance"
-													type="SolarAbsorptance"/>
+												type="SolarAbsorptance"/>
```
nmerket: I've seen these whitespace changes in a couple PRs now. It might be some auto formatting that OxygenXML does. Hopefully this doesn't cause conflicts.

shorowit: Yes. Probably best to apply the Oxygen formatting to master and then merge it into all the PRs to make the whitespace diffs go away.

PR #163, opened 2019-02-07
Presence of Insulation

shorowit:
Documents that `InsulationMaterial/None` implies insulation not present and `InsulationMaterial/Unknown` implies insulation present (but material type unknown).

Also adds a `DuctInsulationMaterial` to allow similar characterization for ducts.

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -168,7 +168,6 @@
 					<xs:documentation>Describe</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element name="None"/>
 			<xs:element name="Unknown"/>
 		</xs:choice>
 	</xs:complexType>
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -168,13 +168,21 @@
 					<xs:documentation>Describe</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element name="Unknown"/>
+			<xs:element name="None">
+				<xs:annotation>
+					<xs:documentation>Indicates insulation not present.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element name="Unknown">
+				<xs:annotation>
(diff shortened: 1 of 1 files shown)


shorowit:
This element is clearer than using the `InsulationMaterial/None` element, so I will remove that.

shorowit:
Move InsulationPresent to Layer element.

shorowit:
I have addressed items 1-3 above. 

I have not addressed item 4 because I feel like it does not need to be spelled out. _Every_ element in HPXML should be considered unknown if not provided.

Any final review/feedback? @GamalielL @lirainer @nmerket

nmerket:
I think that's a reasonable assumption.

GamalielL:
This intersects with the measured vs generated discussion.  Most audit software will have an explicit R-value assumption used in the model and it makes sense to always pass that along.  An interpreter could choose to label the R-value as unknown, if it's not flagged as a user entered value.  We would probably only see the absent R-value in cases where HPXML is strictly used as a means to pass inputs from a GUI to a back end system.  I'm fine with leaving out that documentation.

lirainer:
@shorowit While implementing this I noticed that there are not enumerations in `DuctInsulationMaterial` for some types of insulation that is typically used for ducts. Some options could included (not comprehensive!):
* board - fiberglass, rockwool, ...
* wrap - fiberglass, asbestos, ...
* liner - fiberglass, ...

This overlaps with `DuctMaterial`, so we may want to rethink it. And do you want to keep this as one list with `InsulationMaterial` or have two?

GamalielL:
I've noticed this ambiguity with duct material and insulation before.  We potentially have situations where the duct material is itself insulating (e.g. duct board) and additional insulation is installed over top or even multiple layers of insulation.  We should probable add a full Insulation/Layer system here.  Then we could either have the convention that insulating duct material is described both in the DuctMaterial and in the insulation layers or we could add a separate R-value for the duct material. I'm not sure about adding all of the possible duct insulation materials to the standard list though. That list could become quite cumbersome.

PR #166, opened 2019-02-07
Adds SpecificLeakageArea elements for vented attics/crawls

shorowit:
This allows specifying the SLA for, e.g., vented attics and vented crawlspaces. Note that this branches off the `attics_refactor` branch since it operates on the `Attic` element.


shorowit:
This would be much cleaner if the `AtticType` element had sub-elements rather than being an enumeration. Then we could simply add the `SpecificLeakageArea` underneath the `Attic/Vented` (and the existing `Crawlspace/Vented`) elements. It would also improve consistency between Attics and Foundations. Given that the Attics for v3 are already a breaking change, it seems reasonable to update to this approach.

shorowit:

PR #173, opened 2019-03-28
Adds Refrigerator Type choice

shorowit:
Adds "single door with freezer" as a Refrigerator/Type choice. Available choice in the Weatherization program's software interface. All other types already map into HPXML.

For example:

image

cc @mpathak24

image


nmerket:
My grandparents used to have a refrigerator like that!

PR #178, opened 2019-05-08
Whole slab insulation

shorowit:
Adds an `UnderSlabInsulationSpansEntireSlab` boolean element for the `Slab` to specify that there's whole slab insulation.

image


shorowit:

PR #179, opened 2019-05-09
Foundation wall insulation height

shorowit:
Adds an `InsulationHeight` element for the height of the `FoundationWall/Insulation`. Allows for describing, e.g., 4ft high insulation in an 8ft basement. 

This was implemented similarly to the `PerimeterInsulationDepth` and `UnderSlabInsulationWidth` elements in the `Slab`.

image


shorowit:

Issue #125, opened 2018-07-11
Exterior insulation


DebraLittle:
Should we think about creating a way we can describe exterior insulation via location?  

So we can describe interior/cavity AND exterior. Now, I believe it doesn't accept both. 

CA has required continuous exterior insulation on new homes since Jan 2016. Also for some remodels. 
 Many other jurisdictions also require it.

GamalielL:
It's already possible to have both cavity and continuous insulation layers
on a single surface, but we can't specify one as interior and the other as
exterior.

We have a InsulationLocation element, but unfortunately it's not located in
the Insulation/Layer level. Moving it to the Layer level, where we also
have InstallationType (cavity vs continuous) would solve this.
ᐧ

PR #188, opened 2019-07-26
Allow insulation location to be layer-specific

shorowit:
Addresses #125. Adds "continuous - interior" and "continuous - exterior" to `InstallationType` element, which can be defined for each insulation `Layer`. Removes `InsulationLocation` element. 

This new approach makes it clear that the interior vs exterior insulation location is meant to only apply to continuous insulation layers. The "continuous" option is still allowed if the insulation location is unknown.


shorowit:
1. Add "interior continuous" and "exterior continuous" to InstallationType element.
2. Remove InsulationLocation element.

shorowit:
The items above have been addressed.

PR #189, opened 2019-09-06
Window/Skylight Interior Shading Fraction

shorowit:
For windows/skylights, splits `InteriorShadingFraction` into `InteriorShadingFractionSummer` and `InteriorShadingFractionWinter`. Many software tools and standards use two different numbers.


lirainer:
@nmerket Excuse my git-ignorance, but seems we should note that all these changes have been superseded by #202?

shorowit:
@lirainer What do you mean? The change is preserved in #202, aside from being renamed to `SummerShadingCoefficient` and `WinterShadingCoefficient`:

image

lirainer:
I guess what I'm trying to say is that the changes made in this PR never show up in 3.0 (especially if all the commits are squashed?) It was confusing to me to see the `ShadingFraction` elements show up here when we ended up using `ShadingCoefficient`.

PR #191, opened 2019-09-06
Adding customer comments element

### LINKED ISSUE #170: Homeowner Concerns
Our process places very high priority on Homeowner Concerns. 

By designing projects to solve homeowner concerns, we have greater than 70% closing ratio on $27,600 average job, on a 2300 sf average home. 

Our clients are spending 10-30% of the value of their homes on these projects. It is significantly due to our attention to their concerns. 

Every step in our multi-step process is compensated, so we can avoid conflict of interest caused by pressure to "make a sale" - in other words - no free visits. (In fact we don't even talk to people until they've signed up for a $250-350 "visit."

HVAC contractors are starting to embrace our process. To scale we will need to figure out how to manage the things that matter to the homeowner, not just details about the home.

I imagine HPXML having 5-10 fields for concerns sorted in client priority order. (Common complaints are bad smells, humidity/mold, noise, temperature imbalance, ice damming, etc... The fields shouldn't be predefined as you never know what they will come up with next.)

nmerket:
Fixes #170 

Adds a list of homeowner concerns as `/HPXML/Customer/HomeownerConcerns`

BaseElements_Comments

This would look something like

```xml
<Customer>
    <CustomerDetails>
        <Person>
            <SystemIdentifier id="homeowner"/>
            <Name>Joe Homeowner</Name>
        </Person>
    </CustomerDetails>
    <Comments>
        <Comment>ice damming</Comment>
        <Comment>neighbor's dog poops on my lawn</Comment>
        <Comment>thermal comfort</Comment>
        <Comment>keeping up with Jonses</Comment>
        <!-- can have as many as you need -->
    </Comments>
</Customer>
```

~~I'm not 100% sure this is the best way to do this, so I'd like to have some discussion about it. Maybe this should be called `CustomerConcerns`?~~

Changed to `Comments/Comment` to make more general.

@Tedkidd

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -4902,13 +4902,13 @@ https://hpxml.nrel.gov/wiki/WallTypes</xs:documentation>
 					</xs:sequence>
 				</xs:complexType>
 			</xs:element>
-			<xs:element minOccurs="0" name="HomeownerConcerns">
+			<xs:element minOccurs="0" name="Comments">
 				<xs:annotation>
-					<xs:documentation>A list of homeowner concerns to address in priority order. Examples could include: bad smells, cold or hot rooms, noise, humidity/mold, ice damming.</xs:documentation>
+					<xs:documentation>A list of comments in priority order. </xs:documentation>
 				</xs:annotation>
 				<xs:complexType>
 					<xs:sequence>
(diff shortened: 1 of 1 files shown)


nmerket:
Change homeowner concerns to customer comments.

PR #197, opened 2019-10-02
Fixing project ids

### LINKED ISSUE #85: Project IDs redo
A `Project` node has three possible places to have IDs:
- `Project/BuildingID` - _required_. Why is there only one? Usually projects have two building associated with them.
- `Project/ProjectID` - _optional_. Why is this optional?
- `Project/ProjectDetails/ProjectSystemIdentifiers` - _0 to many_. I'm not really sure what these are for and I've seen different vendors use them differently, so I'm not alone here. Is this to reference the buildings or projects somehow?

These seem to be ambiguous and not really meet the needs of a project. I propose changing this so that projects have a:
- `Project/ProjectID` - _required_, this is an id for the project itself.
- `Project/PreBuildingID` - _required_, this is an id for the pre-retrofit building.
- `Project/PostBuildingID` - _required_, this is an id for the post-retrofit (proposed or actual) building.

Furthermore, we could probably do away with the `Project/ProjectDetails/ProjectStatus` element since it is redundant with the one in the `Building` node.

nmerket:
Fixes #85 by implementing the changes recommended there.

HPXML_Project

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -3763,7 +3763,6 @@ Volume inside the building envelope of the conditioned spaces. This metric can b
 	</xs:complexType>
 	<xs:complexType name="ProjectDetailsType">
 		<xs:sequence>
-			<xs:element maxOccurs="unbounded" name="ProjectSystemIdentifiers" type="RemoteReference"/>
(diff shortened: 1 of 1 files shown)


GamalielL:
@nmerket This looks fine to me.

PR #198, opened 2019-10-02
Home Automation

nmerket:
Fixes #169

PR #200, opened 2019-10-15
DHW Controllers

### LINKED ISSUE #172: Add DHW controllers
DHW timers are becoming a significant factor programs where TOU and demand rates are in use. I wasn't able to find any element that could communicate the presence of these devices.  I propose that we add a DHWController node.  I would put it in the WaterHeating node, at the same level as, but separate from the WaterHeatingSystem node.  It's important for it to be specified separately with it's own SystemIdentifier, so that it can be referenced independently in measures. I haven't learned enough yet about the range of available devices to specify a full list of elements.  Hopefully someone else on the forum has that knowledge.

nmerket:
Fixes #172.

Add the following element under `Building/BuildingDetails/Systems/WaterHeating`.

BaseElements_WaterHeatingControl

It also adds a boolean element `HasMixingValve` to `WaterHeatingSystem`.

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -1397,6 +1397,7 @@
 							<xs:complexType>
 								<xs:sequence>
 									<xs:group ref="SystemInfo"/>
+									<xs:element ref="ConnectedDevice"/>
 									<xs:element minOccurs="0" name="Model" type="xs:string"/>
 									<xs:element minOccurs="0" name="Manufacturer" type="xs:string"/>
 									<xs:element minOccurs="0" name="SerialNumber" type="xs:string"/>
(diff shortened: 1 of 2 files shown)


nmerket:
Note: includes the home automation branch from #198. Probably should be merged after that.

schemas/BaseElements.xsd
```diff
@@ -1486,111 +1314,104 @@
...
-										minOccurs="0">
+									<xs:element name="MeetsACCA5QIHVACSpecification" type="xs:boolean" minOccurs="0"/>
+									<xs:element name="HotWaterTemperature" type="Temperature" minOccurs="0">
 										<xs:annotation>
 											<xs:documentation>[deg F]</xs:documentation>
 										</xs:annotation>
 									</xs:element>
-									<xs:element minOccurs="0" name="UsesDesuperheater"
-										type="xs:boolean">
+									<xs:element minOccurs="0" name="HasMixingValve" type="xs:boolean"/>
```
nmerket: Here's the new `HasMixingValve` element.

schemas/BaseElements.xsd
```diff
@@ -1486,111 +1314,104 @@
...
+									<xs:element ref="ConnectedDevice"/>
+									<xs:element minOccurs="0" name="Model" type="xs:string"/>
+									<xs:element minOccurs="0" name="Manufacturer" type="xs:string"/>
+									<xs:element minOccurs="0" name="SerialNumber" type="xs:string"/>
+									<xs:element minOccurs="0" name="ControlTechnology" type="DHWControllerTechnology"/>
+									<xs:element minOccurs="0" name="TemperatureControl" type="DHWTemperatureControl"/>
+									<xs:element minOccurs="0" ref="extension"/>
+								</xs:sequence>
+							</xs:complexType>
+						</xs:element>
```
nmerket: Here's the new `WaterHeatingControl` element.

schemas/HPXMLDataTypes.xsd
```diff
@@ -2186,4 +2186,28 @@
...
+			<xs:enumeration value="smart"/>
+			<xs:enumeration value="timer"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:simpleType name="DHWTemperatureControl">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="fixed"/>
+			<xs:enumeration value="variable"/>
+		</xs:restriction>
+	</xs:simpleType>
```
nmerket: And the enumerations.

nmerket:
It looks like my editor decided to reformat all the schemas here. There's also the changes from #198. It's making the diffs big. I highlighted the important stuff above.

PR #204, opened 2019-10-18
Allows DuctSurfaceArea to be 0

shorowit:
Previously `DuctSurfaceArea` had to be greater than zero, now it has to be greater than _or equal to_ zero. This allows, for example, describing a ducted evaporative cooler, where you want to declare that there is zero return duct area. (Leaving out a return `Duct` is not the same thing, as it could mean that the return duct information is simply unknown.)


shorowit:
Nevermind, we're going to go a different direction with this.

PR #206, opened 2019-10-23
Percent duct leakage

shorowit:
Allows duct leakage to be entered as percent of airflow.


lirainer:
Is it worth clarifying in the documentation that this is percent of *air flow*? Or is that always assumed?

shorowit:
@lirainer Can you think of any other way that percent duct leakage could be interpreted?

lirainer:
No, I can't. I guess I'm thinking of other ambiguities such as inside/outside or total/supply. But those should be handled elsewhere.

PR #207, opened 2019-10-24
Clarify PV inverter efficiency value

shorowit:
Clarifies that `InverterEfficiency` needs to be 0-1 (fraction), not 0-100 (percentage). I guess this is technically a breaking change if someone was using the percentage convention?


nmerket:
I'm okay with this. It is a subtle but important change.

PR #217, opened 2020-02-11
Additional enumerations for SFA/MF

shorowit:
Adds a few AdjacentTo/Location enumerations useful for single-family attached and multifamily housing. For ERI, this allows mapping as follows:
- Multifamily Buffer Boundary => "other multifamily buffer space" (new, e.g. enclosed unconditioned stairwell)
- Non-Freezing Space => "other non-freezing space" (new, e.g. hared parking garage ceiling)
- Unrated Heated Space => "other heated space" (new, e.g. shared laundry/equipment space)
- Unrated Conditioned Space => "other housing unit" (existing, living/conditioned space of adjacent unit)


schemas/HPXMLDataTypes.xsd
```diff
@@ -2172,6 +2181,12 @@
 			<xs:enumeration value="other exterior"/>
 			<xs:enumeration value="roof deck"/>
 			<xs:enumeration value="unconditioned space"/>
+			<xs:enumeration value="other heated space"/>
```
yzhou601: UnitLocation type is used in `WaterHeatingSystem/Location`, `HVAC/HVACPlant/HeatingSystem | CoolingSystem | HeatPump/UnitLocation`, and `Project/ProjectDetails/Measures/Measure/Location`. I think for these usage, it's probably ok to directly add new enumerations. Does it make sense?

yzhou601:
> @yzhou601 Can you preserve the alphabetical order in these lists? Also, there's no need to include "other housing unit above" and "other housing unit below" in these lists (those are available in the other enumeration because they are needed for the FrameFloor AdjacentTo choices).

Sure, will do.

schemas/HPXMLDataTypes.xsd
shorowit: That seems reasonable.

shorowit:
Now:
- [ ] Document that "other housing unit" refers to the living/conditioned space of the adjacent unit.

Future:
- [ ] Consider a new enumeration like "other housing unit, same space" (or something better)
- [ ] Consider renaming "other housing unit" to be clearer per the description above

nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2023-07-11](https://youtu.be/0vfe3srBLts?t=552)

PR #219, opened 2020-02-21
Compressor type for ACs/HPs

shorowit:
Adds `CompressorType` element for `CoolingSystem` and `HeatPump`. Choices are:
- "single stage"
- "two stage"
- "variable speed"


dhoffmey:
This does seem to be important to add.  Is variable speed anything above two stage?  For example Carrier says...5 stage, variable-speed compression.  If it is commonly understood then fine if not might need to clarify to make explicit what is variable speed.

shorowit:
Good question, @dhoffmey. I will look into this a bit.

shorowit:
I just noticed there's a new [AHRI 2021 Standard for ACs and HPs](https://www.ahrinet.org/Portals/Standards/AHRI%20Standard%20211-241-0B1B-2021%20(SI)%20(1).pdf). The standard characterizes compressors into the 3 above categories only. A couple examples:

image

image

Oddly, it does define a "multiple stage" (or "multiple capacity") system in the definitions section, _but there are no references to this language elsewhere in the standard_. I've asked Jon Winkler to see if he knows anything about this.

Aside from the question of whether we should add a "multiple stage" option, we should also consider renaming "variable speed" to "variable stage" for consistency with the other names.

shorowit:
Information from Jon:

> I did the find the following text by search for “multiple capacity”:
> 3.52. Variable Capacity (or Variable Stage or Variable Speed) System (Variable Stage Air-conditioner or Variable Stage Heat Pump). Air-conditioner(s) or heat pump(s) that has either a Variable Speed Compressor or a Multiple Capacity Compressor.
> 
> I think this means that variable-speed and multiple capacity is being used somewhat interchangeably. For example in Table 8, I would assume multi-stage equipment would need to follow the variable stage system column. 
>

nmerket:
[Discussion in HPXML Working Group meeting 13 Dec 2021](https://youtu.be/MoXAB9W1QA0?t=3286)

PR #221, opened 2020-03-26
Replace `Operable` with `FractionOperable` for Windows/Skylights

shorowit:
Replaces `Window/Operable` (boolean) with `Window/FractionOperable` (0-1).

Since a HPXML `Window` element can represent multiple windows with the same properties (e.g., double pane low-e, vinyl frame, etc.), this change allows describing how much of the window area is operable without needing to split the windows into two HPXML `Window` elements, one for operable windows and one for inoperable windows.

`FractionOperable` description:

> Defined as the sum of the areas for operable windows divided by the sum of the areas for all windows. The fraction should not take into account how much a window can be opened (e.g., 50% for a double hung window). If the Window/Skylight element represents a single window/skylight, the value should be 0 or 1.


nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2022-11-08](https://youtu.be/US-7J1pK0Eo?t=1378)

schemas/BaseElements.xsd
```diff
@@ -4447,7 +4447,16 @@
 						<xs:group ref="SystemInfo"/>
 						<xs:element minOccurs="0" name="GlassType" type="GlassType"/>
 						<xs:element minOccurs="0" name="FrameType" type="WindowFrameType"/>
-						<xs:element minOccurs="0" name="Operable" type="HPXMLBoolean"/>
+						<xs:element minOccurs="0" name="Operable" type="HPXMLBoolean">
+							<xs:annotation>
+								<xs:documentation>DEPRECATED. This will be removed in v4.0. Use FractionOperable element instead.</xs:documentation>
+							</xs:annotation>
+						</xs:element>
```
nmerket, line 4454: Shouldn't we just remove this now if it's going in 4.0?

schemas/BaseElements.xsd
```diff
@@ -4447,7 +4447,16 @@
...
+						<xs:element minOccurs="0" name="FractionOperable" type="HPXMLBoolean">
+							<xs:annotation>
+								<xs:documentation>If the Window/Skylight element represents a single window/skylight, the value should be 0 or 1, otherwise the fraction of area that is operable.</xs:documentation>
```
nmerket, line 4457: Update the description to indicate that it's the sum of the areas of the windows that can open, not the openable area. Use better words.

schemas/BaseElements.xsd
nmerket: Remove the old booleans.

nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2023-09-19](https://youtu.be/CoyZ8DzTGvE?t=1820)

PR #222, opened 2020-03-27
Clarify NumberofConditionedFloorsAboveGrade should include walkout basements

shorowit:
Previously:
- `NumberofConditionedFloors`: Number of floors that are heated/cooled including a basement
- `NumberofConditionedFloorsAboveGrade`: Number of floors above grade that are heated/cooled

Now:
- `NumberofConditionedFloors`: Number of floors that are heated/cooled including a basement
- `NumberofConditionedFloorsAboveGrade`: Number of floors above grade that are heated/cooled  **including a walkout basement**

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -3553,7 +3553,7 @@
 									</xs:element>
 									<xs:element minOccurs="0" name="NumberofConditionedFloorsAboveGrade" type="NumberOfFloorsType">
 										<xs:annotation>
-											<xs:documentation>Number of floors above grade that are heated/cooled</xs:documentation>
+											<xs:documentation>Number of floors above grade that are heated/cooled including a walkout basement</xs:documentation>
 										</xs:annotation>
 									</xs:element>
 									<xs:element name="AverageCeilingHeight" type="LengthMeasurement" minOccurs="0">


shorowit:
Cross-check with other standards (ANSI 301, BPI, 62.2, etc).

GamalielL:
ASHRAE 62.2 uses total above grade conditioned height measured from the lowest above grade point of the pressure envelope to the highest.  So, this element isn't directly relevant. We may want to add an element to capture that measurement. 

When searching the internet I found a lot of indications that walkout basements are not considered above grade floors in the real estate industry.  I think that's worth considering given the efforts to align HPXML with RESO.  Maybe you should check with someone associated with RESO.

shorowit:
Thanks @GamalielL. Indeed here is what [Fannie Mae says](https://selling-guide.fanniemae.com/Selling-Guide/Quick-Reference-Materials/#Gross.20Living.20Area):

> Only finished above-grade areas can be used in calculating and reporting of above-grade room count and square footage for the gross living area. Fannie Mae considers a level to be below-grade if any portion of it is below-grade, regardless of the quality of its finish or the window area of any room. Therefore, a walk-out basement with finished rooms would not be included in the above-grade room count.

shorowit:
Not going to pursue this anymore. Going with the solution in https://github.com/hpxmlwg/hpxml/pull/384 instead.

PR #223, opened 2020-03-27
Adds Dishwasher/ClothesWasher label fields

shorowit:
Adds Energy Guide Label fields for Dishwashers; uses the same names that ClothesWashers use.
- `LabelElectricRate`
- `LabelGasRate`
- `LabelAnnualGasCost`

Also adds `LabelUsage` to both Dishwashers and ClothesWashers, to make a distinction between the cycles/week used in the label/rating and the cycles/week of actual operation (`Usage`).


shorowit:
Update documentation for `Usage` to differentiate with `LabelUsage`.

PR #232, opened 2020-07-15
Shared mech vent system

yzhou601:


schemas/BaseElements.xsd
```diff
@@ -1272,6 +1272,19 @@
...
+														<xs:documentation>[CFM] as rated by manufacturer</xs:documentation>
+													</xs:annotation>
+												</xs:element>
+												<xs:element minOccurs="0" name="SharedSystemFanPower" type="HPXMLDouble">
+													<xs:annotation>
+														<xs:documentation>[W]</xs:documentation>
+													</xs:annotation>
+												</xs:element>
+												<xs:element minOccurs="0" name="FractionOutdoorAir" type="HPXMLDouble"> </xs:element>
+												<xs:element minOccurs="0" name="FractionRecirculation" type="HPXMLDouble"> </xs:element>
```
yzhou601, line 1287: Do we really need both oa and recirc fractions?

schemas/BaseElements.xsd
```diff
@@ -1272,6 +1272,19 @@
 												<xs:element minOccurs="0" name="UsedForWholeBuildingVentilation" type="HPXMLBoolean"/>
 												<xs:element minOccurs="0" name="UsedForSeasonalCoolingLoadReduction" type="HPXMLBoolean"/>
 												<xs:element minOccurs="0" name="UsedForGarageVentilation" type="HPXMLBoolean"/>
+												<xs:element minOccurs="0" name="IsSharedMultifamilySystem" type="HPXMLBoolean"/>
```
shorowit, line 1275: I suggest just calling this `IsSharedSystem` (since it can apply to single-family attached as well), with a note that a shared system is defined as serving multiple building/dwelling units.

schemas/BaseElements.xsd
shorowit, line 1287: I'm pretty sure there is a datatype used specifically for fractions that restricts the values to 0-1. We should use that instead.

shorowit: Not necessarily needed but probably reasonable to have both in case one tool collects one input and another tool collects the other.

shorowit: Should probably also add notes describing these two inputs and declaring that they should sum to 1.

schemas/BaseElements.xsd
shorowit: Here's what I have in another branch: https://github.com/hpxmlwg/hpxml/pull/233/files#diff-45b652e6a591f5d25c2331068b810566R3045-R3049

yzhou601:
@shorowit Base from the shared_hvac seems to be the cleanest way to bring in the datascope attribute.

PR #246, opened 2020-12-15
DSE values as fractions

shorowit:
Forces `AnnualHeatingDistributionSystemEfficiency` and `AnnualCoolingDistributionSystemEfficiency` values to be fractions (0-1). The schema currently allows any number to be provided, so a software tool could be using percentages (0-100). Updated documentation.

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -1183,18 +1183,14 @@
 											<xs:documentation>[sq.ft.] Conditioned floor area that this distribution system serves.</xs:documentation>
 										</xs:annotation>
 									</xs:element>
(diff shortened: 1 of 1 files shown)


shorowit:
@nmerket Not sure if this is considered a breaking change, but it seems important to prevent big inconsistencies between software tools that may be interpreting the values in different ways. Software that was using 0-100 would have to change, while software using 0-1 would not.

nmerket:
This is _technically_ a breaking change as it restricts values that used to be valid. It, however, forces consistency in the way the value is described. I think it's worth bending the rule here to remove the possibility of errors.

PR #249, opened 2021-03-12
Address input vs output heating capacities

shorowit:
For furnaces, [AHRI ratings](https://www.ahridirectory.org/NewSearch?programId=64&searchTypeId=3) include "Input Rating" and "Output Heating Capacity".

Most auditors probably collect input heating capacities. Most BEM tools probably collect output heating capacities. Though HPXML did define `HeatingCapacity` as _input_ heating capacity, it's likely that it was being used inconsistently for different use cases, since there was only a single element available.

To align with AHRI, this PR proposes to:
- Add `HeatingSystem/HeatingInputRating` (and `HeatPump/BackupHeatingInputRating`) elements
- Change the definition of `HeatingSystem/HeatingCapacity`  from _input_ heating capacity to _output_ heating capacity
- Clarify that all `CoolingCapacity` and `HeatPump/HeatingCapacity` and `HeatPump/HeatingCapacity17F` elements are _output_ capacities for consistency

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -3051,9 +3051,14 @@
 				<xs:sequence minOccurs="1" maxOccurs="1">
 					<xs:element name="HeatingSystemType" type="HeatingSystemType" minOccurs="0"/>
 					<xs:element name="HeatingSystemFuel" type="FuelType" minOccurs="0"/>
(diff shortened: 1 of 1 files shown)


shorowit:
General support for this even though it's a _slightly_ breaking change.

PR #257, opened 2021-05-03
Requires AssemblyEffectiveRValue to be > 0

shorowit:
Changes `AssemblyEffectiveRValue` from minInclusive=0 to minExclusive=0. The R-values associated with insulation materials are unchanged (i.e., they can still be zero).

An assembly effective R-value, which is documented as including air films, can never be zero. This change is proposed to prevent software tools from potentially _incorrectly_ using a value of zero. This change helps to ensure that all software tools consistently include air films in their value.


nmerket:
_Technically_ this is a breaking change, but it's relatively minor and has a desired outcome of restricting the values to be realistic, which for most of the known users of HPXML v3 is actually a good thing. So, we're thinking this should go in v3.1 anyway.

PR #264, opened 2021-05-04
Ceilings

### LINKED ISSUE #253: Introduce Ceilings element?
As part of https://github.com/hpxmlwg/hpxml/pull/181, it was debated whether to introduce `Ceilings` (or `FrameCeilings`) as a separate element from `FrameFloors`. It was not included at the time, partly because the difference between ceiling and floor is one of perspective (e.g., the attic floor is the living space's ceiling).

However, there are times where you can't tell whether a `FrameFloor` is a ceiling or floor. The best example is, e.g., InteriorAdjacentTo='living space' and ExteriorAdjacentTo='other housing unit' for a multifamily unit. This surface could be a floor where the other unit is below the current unit, or a ceiling where the other unit is above the current unit. 

Having separate `Ceiling` and `FrameFloor` elements would solve this. 

However, having two elements could also create additional complexity for software developers. Unless additional changes are made, some surfaces could now be described in two ways. For example, the surface between living space and attic could be described as:
- `Ceiling` with InteriorAdjacentTo=living space and ExteriorAdjacentTo=attic, or
- `FrameFloor` with InteriorAdjacentTo=attic and ExteriorAdjacentTo=living space

To address this, we could consider having two different AdjacentTo enumeration lists -- one for InteriorAdjacentTo that includes 'living space' and one for ExteriorAdjacentTo that _excludes_ 'living space'. This would help to ensure that surfaces adjacent to living space are consistently described from the living space perspective.

shorowit:
Closes #253. Introduces a `Ceilings` element, similar to `FrameFloors`. It has all the same child elements except A) any elements with "Floor" in the name are renamed to "Ceiling" and B) the FloorCovering element is dropped since it no longer applies.

Also splits enumerations for `InteriorAdjacentTo`/`ExteriorAdjacentTo` into two sets of enumerations to prevent choices that software tools shouldn't be (and probably aren't) using:
- `InteriorAdjacentTo`: drops "ground", "other housing unit", and "outside"
- `ExteriorAdjacentTo`: drops "living space"

Finally, marks the "other housing unit above" and "other housing unit below" choices as deprecated, since this information is now clear based on whether the surface adjacent to "other housing unit" is a `Ceiling` or `FrameFloor`.

The new `Ceilings` element:

image

### DIFF
--- a/schemas/BaseElements.xsd
+++ b/schemas/BaseElements.xsd
@@ -719,7 +719,7 @@
 								<xs:sequence>
 									<xs:group ref="SystemInfo"/>
 									<xs:element minOccurs="0" ref="AttachedToSpace"/>
-									<xs:element minOccurs="0" name="InteriorAdjacentTo" type="AdjacentTo"/>
+									<xs:element minOccurs="0" name="InteriorAdjacentTo" type="InteriorAdjacentTo"/>
 									<xs:element minOccurs="0" name="Area" type="SurfaceArea">
 										<xs:annotation>
 											<xs:documentation>[sq.ft.] Surface area of the roof itself</xs:documentation>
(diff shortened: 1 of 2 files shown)


shorowit:
Retain `FloorCovering` to allow for wood floor, carpet (for conditioned attics), etc.
Some support for this, but also some concern about software development maintenance.

nmerket:
[Discussion in 7/21/2021 Working Group Meeting](https://youtu.be/rmpzdDNdAlM?t=25m23s)

Issue #292, opened 2021-10-25
Additional `Battery` fields


shorowit:
Non-breaking changes:
- [x] `Quantity`
- [x] `YearInstalled`

Breaking changes:
- [x] capacity fields (e.g., `NominalCapacity`): allow Units/Value child elements to accommodate different units ('kWh' vs 'Ah')

PR #295, opened 2021-11-02
Add battery kWh elements

shorowit:
Adds `NominalCapacitykWh` and `UsableCapacitykWh` elements so we can store capacities in terms of kWh instead of Ah.

Related to https://github.com/hpxmlwg/hpxml/issues/292.


shorowit:
Going to pursue a better breaking change solution for v4.0 instead.

PR #296, opened 2021-11-03
Battery Capacity in kWh

shorowit:
Addresses https://github.com/hpxmlwg/hpxml/issues/292. Allow `Units` and `Value` child elements for battery capacities so that Ah and/or kWh can be used.

image


shorowit:
See if any other units make sense for PowerOutput inputs.

nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2022-07-12](https://youtu.be/AWL1gD9Dh9M?t=59)

PR #307, opened 2022-04-27
Storm Door

yzhou601:


schemas/HPXMLBaseElements.xsd
```diff
@@ -5533,4 +5519,17 @@
...
 		<xs:attribute name="dataSource" type="DataSource"/>
 	</xs:complexType>
+	<xs:complexType name="StormType">
+		<xs:sequence>
+			<xs:group ref="SystemInfo"/>
+			<xs:element minOccurs="0" name="GlazingMaterial" type="GlazingMaterial"/>
+			<xs:element minOccurs="0" name="GlassType" type="GlassType"/>
+			<xs:element minOccurs="0" name="FrameType" type="WindowFrameType"/>
+			<xs:element minOccurs="0" name="Location" type="StormLocation"/>
+			<xs:element minOccurs="0" name="Operable" type="HPXMLBoolean"/>
```
yzhou601, line 5529: @shorowit There's conflict to resolve with v4_0 branch where warning is added and `FractionOperable` is added. I wanted to exclude the deprecated element from `StormDoor`, but it will introduce more complexity that may not worth to do.

shorowit: I wouldn't think that Operable/FractionOperable make sense for a storm door. Isn't a door operable by definition?

yzhou601: I agree, let me try splitting the type and removing `Operable`.

schemas/HPXMLBaseElements.xsd
```diff
@@ -4608,14 +4613,8 @@
 			<xs:element minOccurs="0" name="StormWindow">
 				<xs:complexType>
 					<xs:sequence>
-						<xs:group ref="SystemInfo"/>
-						<xs:element minOccurs="0" name="GlazingMaterial" type="GlazingMaterial"/>
-						<xs:element minOccurs="0" name="GlassType" type="GlassType"/>
-						<xs:element minOccurs="0" name="FrameType" type="WindowFrameType"/>
-						<xs:element minOccurs="0" name="Location" type="StormLocation"/>
+						<xs:group ref="StormInfo"/>
 						<xs:element minOccurs="0" name="Operable" type="HPXMLBoolean"/>
```
shorowit, line 4617: @yzhou601 This will result in a breaking change because the order of elements changes. It's probably fine to use the same elements for StormDoor and StormWindow, even if it's a little weird for `Operable`.

PR #318, opened 2022-06-10
Adds "conditioned space" enumeration

### LINKED ISSUE #242: UnitLocation needs a generic "conditioned space" enumeration
We have "unconditioned space", so it makes sense to have "conditioned space".  The "living space" and "mechanical closet" options are too specific and "other interior" doesn't convey specifically that it's conditioned.  I'm not a fan of growing these lists, so maybe we can establish a convention to handle this.  However, I think "conditioned space" is probably a common designation in software.

shorowit:
Adds "conditioned space" to all element types where "living space" is an enumeration. Closes #242.


nmerket:
🎥 [Discussion in HPXML Working Group meeting 2022-06-24](https://www.youtube.com/watch?v=XQyDFqjQoCM&t=1309s)

Issue #250, opened 2021-03-20
Use xs:assert to improve validation


shorowit:
As of XML Schema 1.1, there is an `<xs:assert>` where you can write XPath expressions to increase validation of elements/values.

There are a number of places this would be useful in HPXML. One example is [HeatingSystem/AnnualHeatingEfficiency/Value](https://hpxml.nrel.gov/datadictionary/3.0.0/Building/BuildingDetails/Systems/HVAC/HVACPlant/HeatingSystem/AnnualHeatingEfficiency/Value) where the only check currently is that the value is > 0 because it handles different units (e.g., AFUE or HSPF). But for some units (e.g., AFUE), we know the value should also be < 1, which could be handled by an xs:assert.

shorowit:
Another example is [WaterHeatingSystem/EnergyFactor](https://hpxml.nrel.gov/datadictionary/3.0.0/Building/BuildingDetails/Systems/WaterHeating/WaterHeatingSystem/EnergyFactor) where you could allow different values for, say, electric storage water heaters vs fossil fuel storage water heaters vs heat pump water heaters.

PR #319, opened 2022-06-21
Disallow zero for efficiencies

shorowit:
Changes the `Efficiency` datatype from `minInclusive=0` to `minExclusive=0`, which will help prevent inappropriate values (whether a bad user input or a software bug). It's hard to imagine how anything could have an efficiency of zero, as that would apply it consumes energy but produces nothing of value.

Also changes `CeilingFan/Efficiency` from an `HPXMLDouble` datatype to the `Efficiency` datatype.

Is this considered a breaking change?

### DIFF
--- a/schemas/HPXMLBaseElements.xsd
+++ b/schemas/HPXMLBaseElements.xsd
@@ -2196,7 +2196,7 @@
 											<xs:documentation>[CFM]</xs:documentation>
 										</xs:annotation>
 									</xs:element>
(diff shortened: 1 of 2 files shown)


nmerket:
Related to #250. Someday we should have better validation across the board for efficiencies.

nmerket:
🎥 [Discussion in HPXML Working Group meeting 2022-06-24](https://www.youtube.com/watch?v=XQyDFqjQoCM&t=452s)

PR #332, opened 2022-07-14
Allow additional floor types

bpark1327:
Fixes #321 

Renames `FrameFloors/FrameFloor` to `Floors/Floor`.
Renames `WallType/StructurallyInsulatedPanel` to `WallType/StructuralInsulatedPanel`
Adds `FloorType` with choices of:
- `WoodFrame`
- `StructuralInsulatedPanel`
- `SteelFrame`
- `SolidConcrete`
- `Other`

PR #351, opened 2022-10-07
Adds `FrameFloor/FloorOrCeiling` element

### LINKED ISSUE #253: Introduce Ceilings element?

shorowit:
Closes #253. Adds `Floor/FloorOrCeiling`: string (choices: "floor", "ceiling"). From the perspective of the living/conditioned space. Often it can be inferred from other information in the HPXML, but not always. This is particularly useful because a lot of standards (like IECC, ANSI 301, etc.) refer to ceilings and floors.

(I'm not calling it `FloorType` because that is being used [here](https://github.com/hpxmlwg/hpxml/pull/332) to support wood frame floors vs mass floors vs ...).

This is a far simpler implementation than [adding an `Enclosure/Ceilings` element](https://github.com/hpxmlwg/hpxml/pull/264), which would require a lot of software updates. 

Should we also deprecate (or mark as deprecated) the "other housing unit above" and "other housing unit below" AdjacentTo enumerations? This is a more flexible implementation and can be used for adjacent locations other than just "other housing unit".

### DIFF
--- a/schemas/HPXMLBaseElements.xsd
+++ b/schemas/HPXMLBaseElements.xsd
@@ -1152,6 +1152,11 @@
 									<xs:element minOccurs="0" ref="AttachedToSpace"/>
 									<xs:element name="ExteriorAdjacentTo" type="AdjacentTo" minOccurs="0"/>
 									<xs:element minOccurs="0" name="InteriorAdjacentTo" type="AdjacentTo"/>
+									<xs:element minOccurs="0" name="FloorOrCeiling" type="FloorOrCeiling">
+										<xs:annotation>
+											<xs:documentation>From the perspective of the living/conditioned space.</xs:documentation>
+										</xs:annotation>
+									</xs:element>
 									<xs:element minOccurs="0" name="FloorJoists" type="StudProperties"/>
 									<xs:element minOccurs="0" name="FloorTrusses" type="StudProperties"/>
 									<xs:element minOccurs="0" name="FloorCovering" type="FloorCovering"/>
--- a/schemas/HPXMLDataTypes.xsd
+++ b/schemas/HPXMLDataTypes.xsd
@@ -4824,4 +4824,17 @@
 			</xs:extension>
 		</xs:simpleContent>
 	</xs:complexType>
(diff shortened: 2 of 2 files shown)


shorowit:
One comment from WG: While we are changing FrameFloor to Floor, rename to a more generic term?

lirainer:
> One comment from WG: While we are changing FrameFloor to Floor, rename to a more generic term?
Others have struggled with this: https://english.stackexchange.com/questions/191958/is-there-a-common-word-for-floor-and-ceiling

nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2022-10-11](https://youtu.be/YSbsZ_VTfHY?t=2721)

RockyMtnBrian:
I like this idea!!

PR #352, opened 2022-10-10
Standalone `Inverter`

### LINKED ISSUE #294: Break out `Inverter` into a standalone component
Currently you specify inverter efficiency on a per `PVSystem` basis. For a home with multiple arrays (say, south facing and west facing), that means having to enter the inverter efficiency twice.

It would be better to have a standalone `Inverter` component that can be referenced by the `PVSystem` components. In that way, a home with multiple arrays could have both `PVSystem`s connected to a single `Inverter` or could have each `PVSystem` connected to its own `Inverter`.

Unfortunately this would be a breaking change.

shorowit:
Closes #294. Moves inverter field into a new `Inverter` element, on the same level as the `PVSystem` element. One or more `PVSystem`s can optionally reference an inverter via a new `AttachedToInverter` element.

Also adds a new `InverterType` element (based on https://powmr.com/blog/types-of-pv-inverters/) with choices of:
- central inverter
- string inverter
- multi-string inverter
- micro inverter
- other

image

### DIFF
--- a/schemas/HPXMLBaseElements.xsd
+++ b/schemas/HPXMLBaseElements.xsd
@@ -1990,14 +1990,12 @@
 										</xs:annotation>
 									</xs:element>
 									<xs:element minOccurs="0" name="NumberOfPanels" type="IntegerGreaterThanZero"/>
-									<xs:element minOccurs="0" name="InverterEfficiency" type="Fraction"/>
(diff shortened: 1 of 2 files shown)


nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2022-11-08](https://youtu.be/US-7J1pK0Eo?t=1060)

PR #366, opened 2023-01-20
Add `LightingGroup/Load` element

shorowit:
Adds a `LightingGroup/Load` element, similar to the `PlugLoad/Load` element. Allows specifying the lighting usage in kWh/yr or W. (There is an existing `LightingGroup/AverageWattage` element, but it's W **per unit** and thus requires the number of units, which may not be known/collected, to obtain the total wattage.)

E.g.:

```xml
<LightingGroup>
    <SystemIdentifier id='LightingGroup'/>
    <Location>interior</Location>
    <Load>
        <Units>kWh/year</Units>
        <Value>620</Value>
    </Load>
</LightingGroup>
```

### DIFF
--- a/merged_schema/HPXMLMerged.xsd
+++ b/merged_schema/HPXMLMerged.xsd
@@ -2334,6 +2334,15 @@
 								<xs:documentation>[sq.ft.]</xs:documentation>
 							</xs:annotation>
 						</xs:element>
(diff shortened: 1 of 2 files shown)


nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2023-05-09](https://youtu.be/62BxnKveGro?t=2160)

PR #368, opened 2023-02-02
Disallow zero for some elements

shorowit:
Found some HPXML files produced by a software tool in which the ventilation fans were described as ERVs with TotalRecoveryEfficiency=0. This doesn't make sense; either TRE needs to be greater than zero or it's actually an HRV.

I looked through all the HPXML elements like these that are of type `Fraction` and propose changing a few of them to use a minimum value of zero _exclusive_ instead of zero _inclusive_:
- VentilationFan/TotalRecoveryEfficiency
- VentilationFan/SensibleRecoveryEfficiency
- VentilationFan/AdjustedTotalRecoveryEfficiency
- VentilationFan/AdjustedSensibleRecoveryEfficiency
- HVACDistribution/AnnualHeatingDistributionSystemEfficiency
- HVACDistribution/AnnualCoolingDistributionSystemEfficiency
- DrainWaterHeatRecovery/Efficiency
- Inverter/InverterEfficiency
- Battery/RoundTripEfficiency

Similar change as what we did for heating, cooling, and ceiling fan efficiencies [here](https://github.com/hpxmlwg/hpxml/pull/319).

### DIFF
--- a/merged_schema/HPXMLMerged.xsd
+++ b/merged_schema/HPXMLMerged.xsd
@@ -1418,12 +1418,12 @@
 											<xs:documentation>[sq.ft.] Conditioned floor area that this distribution system serves.</xs:documentation>
 										</xs:annotation>
 									</xs:element>
-									<xs:element minOccurs="0" name="AnnualHeatingDistributionSystemEfficiency" type="Fraction">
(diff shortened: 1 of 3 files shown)


nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2023-02-14](https://youtu.be/bMDoQKzQnXc?t=1240)

PR #369, opened 2023-02-06
Add combined XSD via automatic commit

### LINKED ISSUE #86: Combine XSD's into a single file
Does anyone object to distributing the HPXML schema as a single xsd instead of the three that we have now? Some platforms and tooling don't do well when having to validate against multiple schemas with `includes`. 

Would that make it harder for anyone out there?

If people agree with this, we could initially distribute a single xsd in addition to the three xsd's which wouldn't be a breaking change. Eventually we could distribute only the single xsd during one of the major breaking changes.

shorowit:
Closes #86.

GitHub CI automatically merges all XSDs into a single `merged_schema/HPXMLMerged.xsd`, commits it to the repo, and verifies that an example HPXML validates against it.

Also adds an `examples/invalid.xml` for testing purposes.


scripts/xsdflatten.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python
+import sys
+import re
+import copy
+from lxml import etree
+
+
+def get_includes_from_file(filename):
+    pattern = re.compile('(<xs:include schemaLocation)')
```
nmerket, line 9: `<xs:include/>` can have [other attributes](https://www.w3schools.com/xml/el_include.asp). If one of those slides in before the `schemaLocation`, it could be missed. 

```xml
<xs:include id="something" schemaLocation="BaseElements.xsd">
```

I'd suggest using an xpath to find these instead:

```xpath
//xs:include/@schemaLocation
```

shorowit: Good call. For what it's worth, this code came from [here](https://github.com/esunder/xsdflatten).

nmerket (approved):

nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2023-02-14](https://youtu.be/bMDoQKzQnXc?t=1104)

PR #371, opened 2023-02-13
`DetailedConstruction` element for surfaces

shorowit:
Strawman proposal:

HPXMLBaseElements_DetailedConstruction

Initial notes/questions:
- Documentation calls out that layers should be ordered from exterior to interior.
- Documentation calls out that air films should not be included.
- Should each construction layer have its own SystemIdentifier?
- Should we include a standardized list of common materials? (We would still need to allow a user-specified value.)
- Should we allow for parallel paths for, e.g., surfaces with frame/cavity construction? Note: For some wall types, parallel path calculations are not appropriate. For metal frame walls, ASHRAE provides correction factors.
  - Some software tools describe the insulation grade using a path. Should insulation grading be included or excluded? There is a separate [InsulationGrade](https://hpxml.nrel.gov/datadictionary/3.1.0/Building/BuildingDetails/Enclosure/Walls/Wall/Insulation/InsulationGrade) element.
- What is relationship between these inputs and existing inputs (insulation layers, framing factor, siding, assembly R-value, etc.)?

### DIFF
--- a/schemas/HPXMLBaseElements.xsd
+++ b/schemas/HPXMLBaseElements.xsd
@@ -186,6 +186,29 @@
 		</xs:sequence>
 		<xs:attribute name="dataSource" type="DataSource"/>
 	</xs:complexType>
+	<xs:complexType name="DetailedConstruction">
+		<xs:sequence>
+			<xs:group ref="SystemInfo"/>
+			<xs:element maxOccurs="unbounded" name="ConstructionLayer">
+				<xs:annotation>
+					<xs:documentation>Layers should be ordered from exterior to interior. Do not include air films.</xs:documentation>
+				</xs:annotation>
+				<xs:complexType>
+					<xs:sequence>
+						<xs:element minOccurs="0" name="Material" type="xs:string"/>
+						<xs:element minOccurs="0" name="Thickness" type="HPXMLDouble"/>
+						<xs:element minOccurs="0" name="Conductivity" type="HPXMLDouble"/>
+						<xs:element minOccurs="0" name="Density" type="HPXMLDouble"/>
+						<xs:element minOccurs="0" name="SpecificHeat" type="HPXMLDouble"/>
+						<xs:element minOccurs="0" name="RValue" type="HPXMLDouble"/>
+						<xs:element ref="extension" minOccurs="0"/>
+					</xs:sequence>
+				</xs:complexType>
+			</xs:element>
+			<xs:element ref="extension" minOccurs="0"/>
+		</xs:sequence>
+		<xs:attribute name="dataSource" type="DataSource"/>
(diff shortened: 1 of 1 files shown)


shorowit:
- Add `LayerType` element with choices like "siding", "insulation", "sheathing", "interior finish", ...

nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2023-02-14](https://youtu.be/bMDoQKzQnXc?t=1653)

merged_schema/HPXMLMerged.xsd
```diff
@@ -200,6 +200,64 @@
...
+										<xs:annotation>
+											<xs:documentation>[Btu/lb-F]</xs:documentation>
+										</xs:annotation>
+									</xs:element>
+									<xs:element minOccurs="0" name="RValue" type="RValue"/>
+									<xs:element ref="extension" minOccurs="0"/>
+								</xs:sequence>
+							</xs:complexType>
+						</xs:element>
+					</xs:sequence>
```
nmerket, line 254: Suggest putting an `extension` after `LayerMaterial`. 

```suggestion
						</xs:element>
						<xs:element ref="extension" minOccurs="0"/>
					</xs:sequence>
```

schemas/HPXMLBaseElements.xsd
```diff
@@ -186,6 +186,64 @@
...
+										<xs:annotation>
+											<xs:documentation>[Btu/lb-F]</xs:documentation>
+										</xs:annotation>
+									</xs:element>
+									<xs:element minOccurs="0" name="RValue" type="RValue"/>
+									<xs:element ref="extension" minOccurs="0"/>
+								</xs:sequence>
+							</xs:complexType>
+						</xs:element>
+					</xs:sequence>
```
nmerket, line 240: Suggest putting an `extension` after `LayerMaterial`. 

```suggestion
						</xs:element>
						<xs:element ref="extension" minOccurs="0"/>
					</xs:sequence>
```

shorowit:
- Top to bottom for horizontal surfaces instead?
- What about interior partition walls/floors (e.g., floor between conditioned stories)?
- Think about expanding the LayerType choices.

nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2023-03-14](https://youtu.be/alTvyoUs2qk?t=265)

shorowit:
@GamalielL Regarding the convention of the order of construction layers, what do you think about an element to describe it? For example, there could be a `ConstructionLayersOrder` element with choices of "exterior to interior", "interior to exterior", "top to bottom", and "bottom to top". Is the flexibility worth the extra complexity? Or should we just force all HPXML users to do it the same way?

PR #375, opened 2023-03-13
Run CI on master branch too

shorowit:


nmerket (approved):
Ooooh. It's because the merged file got updated. Yeah, makes sense.

PR #381, opened 2023-04-10
Fix broken link in README

nelsonauner:
First, thanks for supporting these fantastic tools!

When perusing documentation I noticed a broken link in the README. 
`http://hpxmlonline.com` is broken because the server is set to reject non-HTTPS requests instead of redirecting them. 
So, a user clicking this link in the readme will, for most browsers, wait 30 seconds and then receive a time out error.

The best way to fix this is to add logic to your webserver to redirect non-secure requests to HTTPS, but changing the links is also useful. 

Note that I have not changed any of the non-secure links in the XML namespace (e.g. `<HPXML xmlns="https://hpxmlonline.com/2019/10"  schemaVersion="3.1">` because these uses are just namespaces, not actual resolveable URLs, and are sensitive to the protocol (e.g. `http` vs `https`)

https://stackoverflow.com/questions/30707609/xml-namespace-uri-with-https

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ HPXML
 
 Home Performance XML (HPXML) is a data transfer standard for the home performance industry. This repository is where the development of the schemas happens. 
 
-* `Official HPXML Website <http://hpxmlonline.com>`_ - includes press releases, and general information about the project.
+* `Official HPXML Website <https://hpxmlonline.com>`_ - includes press releases, and general information about the project.
 * `HPXML Guide <http://hpxml-guide.readthedocs.org/en/latest/>`_ - Draft documentation including program administrator and software implementer guides. Includes examples and explanation of how to use HPXML.
 * `HPXML Toolbox <https://hpxml.nrel.gov>`_ - Validator and Data Dictionary (schema explorer).
 
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ HPXML
 
 Home Performance XML (HPXML) is a data transfer standard for the home performance industry. This repository is where the development of the schemas happens. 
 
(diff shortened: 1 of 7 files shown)


nelsonauner:
@nmerket @shorowit I didn't see a `CONTRIBUTING.md` or anything on how to contribute so gave it my best shot :)

shorowit:
Both http and https immediately resolve for me, but it appears that the www is now needed.

http://www.hpxmlonline.com
https://www.hpxmlonline.com

nelsonauner:
@nmerket You closed this pull request, but the current `master` branch is broken, in that https://github.com/hpxmlwg/hpxml/blame/master/README.rst has a link to `http://hpxmlonline.com` and that link does not resolve. 

If per @shorowit 's suggestion you'd rather have `http://www.hpxmlonline.com`  instead https, that's fine, and I'm happy to change the PR! But there is something broken currently

nmerket:
Oh I see. Yeah, let's change that to **www**.hpxmlonline.com. Thanks again.

nelsonauner:
ok! @nmerket 

I took your and scotts comment to mean that you preferred to fix issue this issue by specifying subdomain (`www`) and keeping non-secure scheme (`http://`) over `https://`.

I really love y'alls work and would be happy to help draft a `CONTRIBUTING.md` or whatever to minimize annoying drive-bys like this one. LMK!

shorowit:
@nmerket The link under About also needs to be updated:

image

nelsonauner:
Good catch Scott! 

LMK if anything else I should do (e.g. rebase into a single commit before merge)

nmerket:
You're good. I'll take it from here. Thank for pointing this out.

nmerket (approved):

PR #386, opened 2023-07-21
Remove redundant datatypes

### LINKED ISSUE #339: Remove redundant datatypes
There are probably quite a few datatypes in HPXMLDataTypes.xsd that are essentially equivalent. It would be nice to clean it up so that the file can be significantly simplified and reduced in size.

For example:
```
	<xs:simpleType name="CAZDepressurizationLimit_simple">
		<xs:restriction base="xs:double"/>
	</xs:simpleType>
	<xs:complexType name="CAZDepressurizationLimit">
		<xs:simpleContent>
			<xs:extension base="CAZDepressurizationLimit_simple">
			</xs:extension>
		</xs:simpleContent>
	</xs:complexType>
```

Any references to this datatype could simply use the generic `HPXMLDouble` instead.

shorowit:
Closes #339. Removes 88 redundant datatypes in favor of more generic datatypes (e.g., `HPXMLDoubleGreaterThanZero`). Non-breaking change; everything should behave the same as before.

### DIFF
--- a/schemas/HPXMLBaseElements.xsd
+++ b/schemas/HPXMLBaseElements.xsd
@@ -1411,12 +1411,12 @@
 											<xs:documentation>[sq.ft.] Conditioned floor area that this distribution system serves.</xs:documentation>
 										</xs:annotation>
 									</xs:element>
-									<xs:element minOccurs="0" name="AnnualHeatingDistributionSystemEfficiency" type="HPXMLFractionExcludingZero">
+									<xs:element minOccurs="0" name="AnnualHeatingDistributionSystemEfficiency" type="HPXMLFractionGreaterThanZero">
 										<xs:annotation>
 											<xs:documentation>For software that does not calculate an annual distribution system efficiency (DSE) for heating, the DSE may be approximated by equation 3.4.i in ANSI/BPI-2400-S-2012: Standard Practice for Standardized Qualification of Whole-House Energy Savings, Predictions by Calibration to Energy Use History. Enter values as a fractional number between 0 and 1, i.e. 80% = 0.8</xs:documentation>
 										</xs:annotation>
 									</xs:element>
-									<xs:element minOccurs="0" name="AnnualCoolingDistributionSystemEfficiency" type="HPXMLFractionExcludingZero">
+									<xs:element minOccurs="0" name="AnnualCoolingDistributionSystemEfficiency" type="HPXMLFractionGreaterThanZero">
 										<xs:annotation>
 											<xs:documentation>For software that does not calculate an annual distribution system efficiency (DSE) for cooling, the DSE may be approximated by equation 3.4.i in ANSI/BPI-2400-S-2012: Standard Practice for Standardized Qualification of Whole-House Energy Savings, Predictions by Calibration to Energy Use History. Enter values as a fractional number between 0 and 1, i.e. 80% = 0.8</xs:documentation>
 										</xs:annotation>
@@ -1542,7 +1542,7 @@
 														<xs:documentation>[sones] as tested in field</xs:documentation>
 													</xs:annotation>
 												</xs:element>
-												<xs:element minOccurs="0" name="TotalRecoveryEfficiency" type="HPXMLFractionExcludingZero">
+												<xs:element minOccurs="0" name="TotalRecoveryEfficiency" type="HPXMLFractionGreaterThanZero">
 													<xs:annotation>
 														<xs:documentation>The net total energy (sensible plus latent, also called enthalpy) recovered by the supply airstream adjusted by electric
 															consumption, case heat loss or heat gain, air leakage and airflow mass imbalance between the two airstreams, as a percent of the potential
@@ -1550,7 +1550,7 @@
 															(hvi.org).</xs:documentation>
 													</xs:annotation>
 												</xs:element>
-												<xs:element minOccurs="0" name="SensibleRecoveryEfficiency" type="HPXMLFractionExcludingZero">
+												<xs:element minOccurs="0" name="SensibleRecoveryEfficiency" type="HPXMLFractionGreaterThanZero">
 													<xs:annotation>
 														<xs:documentation>The net sensible energy recovered by the supply airstream as adjusted by electric consumption, case heat loss or heat gain,
 															air leakage, airflow mass imbalance between the two airstreams and the energy used for defrost (when running the Very Low Temperature Test),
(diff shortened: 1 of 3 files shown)


schemas/HPXMLBaseElements.xsd
```diff
@@ -2194,30 +2194,30 @@
...
+									<xs:element minOccurs="0" name="Manufacturer" type="HPXMLString"/>
+									<xs:element maxOccurs="1" minOccurs="0" name="ModelNumber" type="HPXMLString"/>
 									<xs:element minOccurs="0" name="SerialNumber" type="HPXMLString"/>
 									<xs:element minOccurs="0" ref="ConnectedDevice"/>
 									<xs:element minOccurs="0" name="ChargingLevel" type="EVChargingLevel"/>
 									<xs:element minOccurs="0" name="ChargingConnector" type="EVChargingConnector"/>
-									<xs:element maxOccurs="1" minOccurs="0" name="ModelYear" type="Year"/>
-									<xs:element minOccurs="0" name="ACPowerSourceVoltage" type="Voltage">
+									<xs:element maxOccurs="1" minOccurs="0" name="ModelYear" type="HPXMLInteger"/>
+									<xs:element minOccurs="0" name="ACPowerSourceVoltage" type="HPXMLDecimalGreaterThanZero">
```
shorowit, line 2205: There are a few elements, like this one, that use decimal. Why not double?

schemas/HPXMLBaseElements.xsd
```diff
@@ -142,11 +142,11 @@
 			<xs:element name="Name">
 				<xs:complexType>
 					<xs:sequence>
-						<xs:element name="PrefixName" type="PrefixName" minOccurs="0"/>
+						<xs:element name="PrefixName" type="HPXMLString" minOccurs="0"/>
```
shorowit, line 145: Some of these changes seem like no-brainers.

schemas/HPXMLBaseElements.xsd
```diff
@@ -401,7 +401,7 @@
 					</xs:element>
 					<xs:element maxOccurs="unbounded" minOccurs="0" name="ControlType" type="ClothesDryerControlType"/>
 					<xs:element minOccurs="0" name="Vented" type="HPXMLBoolean"/>
-					<xs:element minOccurs="0" name="VentedFlowRate" type="FlowRate"/>
+					<xs:element minOccurs="0" name="VentedFlowRate" type="HPXMLDoubleGreaterThanOrEqualToZero"/>
```
shorowit, line 404: There are pros and cons to replacing these kinds of datatypes (referenced by multiple elements):
- **Pros**: It's now much easier to see directly what the datatype and restrictions are; they are no longer obscured. In fact, I have found a couple questionable datatypes being assigned but it wasn't easy to tell before. There's also less confusion about when to use a generic datatype (like `HPXMLDouble`) vs a more specific datatype (like `FlowRated`).
- **Cons**: When adding, e.g., a new flow rate element in the future, the developer won't be able to simply choose an existing `FlowRate` datatype, rather they'd probably want to look through the schema for similar elements and choose the same datatype. So it requires a bit more brain power.

If the cons outweights the pros, we can consider retaining datatypes like `FlowRate`, `Capacity`, `RValue`, etc., which are reference by multiple elements.

schemas/HPXMLBaseElements.xsd
```diff
@@ -5653,10 +5653,10 @@
 	<xs:complexType name="WaterHeaterImprovementInfo">
 		<xs:sequence>
 			<xs:element name="JacketInstalledIndicator" type="HPXMLBoolean" minOccurs="0"/>
-			<xs:element name="DispositionofExistingSystem" type="DispositionofExistingSystem" minOccurs="0"/>
+			<xs:element name="DispositionofExistingSystem" type="HPXMLString" minOccurs="0"/>
 			<xs:element name="RepairsDescription" type="HPXMLString" minOccurs="0" maxOccurs="unbounded"/>
-			<xs:element name="PipeInsulated" type="PipeInsulated" minOccurs="0"/>
-			<xs:element name="LengthofPipeInsulated" type="LengthMeasurement" minOccurs="0">
+			<xs:element name="PipeInsulated" type="HPXMLString" minOccurs="0"/>
```
shorowit, line 5658: Seems like this should be a boolean?

schemas/HPXMLBaseElements.xsd
```diff
@@ -4482,15 +4482,15 @@
...
+									<xs:element name="LoadProfile" type="HPXMLString" minOccurs="0">
 										<xs:annotation>
 											<xs:documentation>A load profile is created using measurements of a customer's electricity use at regular intervals, typically one hour or less, and
 												provides an accurate representation of a customer's usage pattern over time.</xs:documentation>
 										</xs:annotation>
 									</xs:element>
-									<xs:element name="Quantity" type="Quantity"/>
-									<xs:element name="AnnualAmount" type="AnnualAmount" minOccurs="0"/>
+									<xs:element name="Quantity" type="HPXMLDecimal"/>
+									<xs:element name="AnnualAmount" type="HPXMLString" minOccurs="0"/>
```
shorowit, line 4493: Seems like this should be a double?

schemas/HPXMLBaseElements.xsd
```diff
@@ -4452,21 +4452,21 @@
...
+						<xs:element name="Value" type="HPXMLDecimal"/>
 					</xs:sequence>
 					<xs:attribute name="dataSource" type="DataSource"/>
 				</xs:complexType>
 			</xs:element>
 			<xs:element name="Location" type="UnitLocation" minOccurs="0"/>
-			<xs:element name="EstimatedLife" type="EstimatedLife" minOccurs="0"/>
-			<xs:element name="InstallationDate" type="InstallationDate" minOccurs="0"/>
-			<xs:element name="Cost" type="Cost" minOccurs="0"/>
+			<xs:element name="EstimatedLife" type="HPXMLString" minOccurs="0"/>
```
shorowit, line 4467: Should this be a double? No idea how this is used.

schemas/HPXMLBaseElements.xsd
shorowit: I did notice an example where the use of `RValue` for different elements is arguably not appropriate, but the issue was obscured by just pointing to an `RValue` datatape.
- When referring to, say, the perimeter insulation R-value for a slab, allowing zero is fine.
- When referring to a door, an R-value must be greater than zero.

PR #394, opened 2023-10-02
Fix RemoteReference documentation

shorowit:
Should have been updated as part of https://github.com/hpxmlwg/hpxml/pull/378


nmerket (approved):
Good catch.

PR #229, opened 2020-05-29
Add PermanentSpa and PortableSpa

joseph-robertson:
There are now 3 elements:

# Pools

This is mostly unchanged except that it moves all the subelements to a type to be reused.

**Breaking Change** `Pools/Pool/PoolPumps/PoolPump` have been renamed to `Pools/Pool/Pumps/Pump`. This is for consistency in the PermanentSpas element.

HPXMLBaseElements_Pools

# Spas

There is now a `Spas` element next to `Pools` that can contain 0 or more `PermanentSpa` or `PortableSpa` subelements. 

- `PermanentSpa` elements are very similar to `Pool` and extend the same complex data type. Additional spa-specific elements can be added with this type extension organization.
- `PortableElectricSpa` has the elements in the Energy Guide label below. The data field annotations reference the standard. 

HPXMLBaseElements_Spas

PR #395, opened 2023-10-02
Portable Electric Spa Fixes

nmerket:
Related to #229.

Based on feedback from Jennifer Hatfield with PHTA:

- Changes the datatype for `TotalAnnualPowerConsumptionInStandbyMode` to be `RatedAnnualkWh`, which is accurately just a single number. The previous data type selection was an error.
- Removes `FilterType` from `PortableElectricSpa` because they all use cartridge filters, so there is no good reason to ask for something that will likely lead to an erroneous input. 

HPXMLBaseElements_PortableElectricSpa


shorowit (approved):
Looks reasonable to me.

PR #399, opened 2024-01-19
SystemIdentifier attributes

shorowit:
Replaces the `sameas` attribute with `retrofit_id` to better clarify how the attribute is intended to be used. The attribute description is: _"Use to reference the id of the matching object on the base (pre-retrofit) building where the object has not been replaced."_

Adds a new `shared_id` attribute with the description: _"Use to reference the id of the same object on another multifamily unit (e.g., a single shared boiler or common wall described in multiple multifamily units)."_


shorowit:
Some software may be using `sameas` even for situations where the element (e.g., HVAC system) is replaced. Concerns about the "where the object has not been replaced" part of the description; should that be stricken?

Alternatives to `pre_retrofit_id`... `base_id` or `upgraded_id`?

Needs further discussion.

nmerket:
:movie_camera: [Discussion in HPXML Working Group Meeting 2024-03-14](https://youtu.be/INv483NYz4M?t=2534)

shorowit:
> Some software may be using `sameas` even for situations where the element (e.g., HVAC system) is replaced, or for an AC/furnace->HP situation. Concerns about the "_where the object has not been replaced_" part of the description; should that be stricken?

It seems like this attribute has been misunderstood/misused. Indeed, even I thought the `sameas` attribute was supposed to be used for, e.g., a wall that has been upgraded during a retrofit, but it turns out that's not the case. It's meant to describe components that **do not change** during a retrofit. When something changes (either a wall upgrade or an HVAC replacement), a measure is supposed to be defined that links the before and after components. See these sections of the HPXML Implementation Guide:
- https://hpxml-guide.readthedocs.io/en/latest/software_developer/references.html#sameas
- https://hpxml-guide.readthedocs.io/en/latest/software_developer/usecases/auditupgrade.html#post-upgrade

(It's also not possible to use `sameas` to describe an AC/furnace->HP replacement since the HP can only refer to one pre-retrofit component.)

shorowit:
Replaced by #437.

PR #403, opened 2024-02-12
Vehicles

### LINKED ISSUE #205: Electric Vehicle
We already have EV chargers, why not the vehicle itself? 

Include whether it is communicating or not, etc. 

What else should this have?

@Tedkidd

aspeake:
Closes #205.

Introduces a new Vehicles object with the ability to define various vehicle types. Additionally for EVs, enables referencing an ElectricVehicleCharger id, and adds an optional `EVChargerLocation` field for ElectricVehicleCharger. There are a set of new elements for vehicles that apply to any type, and five unique vehicle types, ICE, BEV, PHEV, HEV, and Other. Each type has elements that apply to only subset of types (eg, fuel type for ICE, battery capacity for BEV).

HPXMLBaseElements_Vehicles

HPXMLBaseElements_BEV


schemas/HPXMLBaseElements.xsd
```diff
@@ -6268,6 +6269,11 @@
 												in to the energy retrieved from storage.</xs:documentation>
 				</xs:annotation>
 			</xs:element>
+			<xs:element minOccurs="0" name="FractionChargedAtHome" type="Fraction">
+				<xs:annotation>
+					<xs:documentation>The raction of charging energy provided by an at-home charger.</xs:documentation>
```
shorowit, line 6274: @aspeake "raction" typo

merged_schema/HPXMLMerged.xsd
```diff
@@ -2216,6 +2216,110 @@
...
+					<xs:sequence>
+						<xs:element maxOccurs="unbounded" name="Vehicle">
+							<xs:complexType>
+								<xs:sequence>
+									<xs:group ref="SystemInfo"/>
+									<xs:element minOccurs="0" name="Manufacturer" type="Manufacturer"/>
+									<xs:element minOccurs="0" name="Model" type="Model"/>
+									<xs:element minOccurs="0" name="ModelYear" type="Year"/>
+									<xs:element minOccurs="0" name="VehicleClass" type="VehicleClassType"/>
+									<xs:element minOccurs="0" name="VehicleWeight" type="HPXMLDoubleGreaterThanZero"/>
```
nmerket, line 2230: What are the units?

merged_schema/HPXMLMerged.xsd
```diff
@@ -2216,6 +2216,110 @@
...
+													<xs:complexType>
+														<xs:sequence>
+															<xs:element minOccurs="0" name="Battery" type="VehicleBattery"/>
+															<xs:element minOccurs="0" name="FractionChargedAtHome" type="Fraction">
+																<xs:annotation>
+																	<xs:documentation>The fraction of charging energy provided by an at-home charger.</xs:documentation>
+																</xs:annotation>
+															</xs:element>
+															<xs:element minOccurs="0" name="ConnectedCharger" type="LocalReference"/>
+															<xs:element minOccurs="0" name="FuelType" type="VehicleFuelType"/>
```
nmerket, line 2268: PHEV fuel type is a little confusing. Is it gas or electric? I assume this means the part of the fuel that isn't electricity. Should we just remove "electricity" from the fuel type enumerations? If you have a BEV, there isn't a fuel type field and if you have one of the others, it would be wrong to say electricity, right?

merged_schema/HPXMLMerged.xsd
```diff
@@ -2216,6 +2216,110 @@
...
+														</xs:sequence>
+													</xs:complexType>
+												</xs:element>
+											</xs:choice>
+										</xs:complexType>
+									</xs:element>
+									<xs:element minOccurs="0" name="PrimaryLocation" type="VehicleLocation"/>
+									<xs:element minOccurs="0" name="MilesDrivenPerYear" type="HPXMLDoubleGreaterThanZero"/>
+									<xs:element minOccurs="0" name="HoursDrivenPerWeek" type="HPXMLDoubleGreaterThanOrEqualToZero"/>
+									<xs:element minOccurs="0" name="FuelEconomy">
```
nmerket, line 2305: Maybe we should allow more than one of these like we do for heating and cooling efficiency in case there are more than one way to describe this for the same car. 

Is this the EPA rating? I would encourage you to look at the [EPA fuel economy label](https://www.epa.gov/greenvehicles/learn-about-fuel-economy-label) and have those fields in here.

merged_schema/HPXMLMerged.xsd
aspeake: Good point, I removed electricity as an option and removed FuelType from BEV

aspeake:
- Introduce V2x fields in VehicleBattery
- FractionChargedAtHome --> allow home, work, public, other that adds up to 100%
- Potential new fields:
  - VIN
  - Trim level
  - Drive system 
  - mileage
  - extension elements under each vehicle type

aspeake:
Latest updates:
- `VehicleBattery` type fields for `GridConnected` and `BuildingConnected`
- extension element for each vehicle type and the `VehicleBattery` type
- New `Vehicle` elements:
  - `Vehicle/VIN`
  - `Vehicle/ModelTrim`
  - `Vehicle/VehicleClass`: 1-13 per U.S. FHA
  - `Vehicle/Mileage`
  - `FuelEconomyCombined` field with `Units` and `Value`
  - `FuelEconomyCity` field with `Units` and `Value`
  - `FuelEconomyHighway` field with `Units` and `Value`
-  `FractionChargedLocation` elements for BEV and PHEV
   -  `Location`
   - `Percentage`

aspeake:
WG Comments:
- Look into introducing maximum charging rate on the vehicle or panel
-  Expand description of `ElectricVehicleCharger` and/or `ConnectedCharger` with examples of what it can describe (EVSE, level 2 chargers, etc)

schemas/HPXMLDataTypes.xsd
```diff
@@ -3715,6 +3727,94 @@
...
+			<xs:extension base="VehicleFuelType_simple"/>
+		</xs:simpleContent>
+	</xs:complexType>
+	<xs:simpleType name="VehicleClassType_simple">
+		<xs:annotation>
+			<xs:documentation>The class as defined by the U.S. Federal Highway Administration</xs:documentation>
+		</xs:annotation>
+		<xs:restriction base="xs:int">
+			<xs:minExclusive value="1"/>
+			<xs:maxExclusive value="13"/>
```
shorowit, line 3796: This doesn't look right. Should these be Inclusive instead of Exclusive? Currently it allows values 2-12.

shorowit: That said, I find it weird to use integers here. It would be far more user-friendly to use string enumerations (motorcycle, passenger car, etc.)

schemas/HPXMLDataTypes.xsd
```diff
@@ -3715,6 +3727,94 @@
...
+			<xs:enumeration value="PHEV"/>
+			<xs:enumeration value="HEV"/>
+			<xs:enumeration value="ICE"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:complexType name="VehicleType">
+		<xs:simpleContent>
+			<xs:extension base="VehicleType_simple"/>
+		</xs:simpleContent>
+	</xs:complexType>
```
shorowit, line 3742: No longer used, right?

schemas/HPXMLDataTypes.xsd
```diff
@@ -3715,6 +3727,94 @@
...
+			<xs:enumeration value="Work"/>
+			<xs:enumeration value="Public"/>
+			<xs:enumeration value="Other"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:complexType name="VehicleChargingLocation">
+		<xs:simpleContent>
+			<xs:extension base="VehicleChargingLocation_simple"/>
+		</xs:simpleContent>
+	</xs:complexType>
```
shorowit, line 3755: Needs the dataSource attribute.

schemas/HPXMLDataTypes.xsd
```diff
@@ -3715,6 +3727,94 @@
...
+			<xs:enumeration value="unconditioned space"/>
+			<xs:enumeration value="other heated space"/>
+			<xs:enumeration value="other non-freezing space"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:complexType name="VehicleLocation">
+		<xs:simpleContent>
+			<xs:extension base="VehicleLocation_simple"/>
+		</xs:simpleContent>
+	</xs:complexType>
```
shorowit, line 3773: Needs the dataSource attribute.

schemas/HPXMLDataTypes.xsd
```diff
@@ -3715,6 +3727,94 @@
...
+			<xs:enumeration value="hydrogen"/>
+			<xs:enumeration value="natural gas"/>
+			<xs:enumeration value="propane"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:complexType name="VehicleFuelType">
+		<xs:simpleContent>
+			<xs:extension base="VehicleFuelType_simple"/>
+		</xs:simpleContent>
+	</xs:complexType>
```
shorowit, line 3789: Needs the dataSource attribute.

schemas/HPXMLDataTypes.xsd
```diff
@@ -3715,6 +3727,94 @@
...
+		<xs:restriction base="xs:int">
+			<xs:minExclusive value="1"/>
+			<xs:maxExclusive value="13"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:complexType name="VehicleClassType">
+		<xs:simpleContent>
+			<xs:extension base="VehicleClassType_simple"/>
+		</xs:simpleContent>
+	</xs:complexType>
```
shorowit, line 3803: Needs the dataSource attribute.

schemas/HPXMLDataTypes.xsd
```diff
@@ -3715,6 +3727,94 @@
...
+			<xs:enumeration value="km/L"/>
+			<xs:enumeration value="mile/kWh"/>
+			<xs:enumeration value="kWh/mile"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:complexType name="FuelEconomyUnits">
+		<xs:simpleContent>
+			<xs:extension base="FuelEconomyUnits_simple"/>
+		</xs:simpleContent>
+	</xs:complexType>
```
shorowit, line 3817: Needs the dataSource attribute.

schemas/HPXMLBaseElements.xsd
```diff
@@ -2248,6 +2248,170 @@
...
+															<xs:element minOccurs="0" name="FuelType" type="VehicleFuelType"/>
+															<xs:element ref="extension" minOccurs="0"/>
+														</xs:sequence>
+													</xs:complexType>
+												</xs:element>
+											</xs:choice>
+										</xs:complexType>
+									</xs:element>
+									<xs:element minOccurs="0" name="PrimaryLocation" type="VehicleLocation"/>
+									<xs:element minOccurs="0" name="Mileage" type="HPXMLDoubleGreaterThanOrEqualToZero"/>
```
shorowit, line 2376: Disallow zero?

schemas/HPXMLBaseElements.xsd
```diff
@@ -6230,6 +6395,56 @@
 		</xs:sequence>
 		<xs:attribute name="dataSource" type="DataSource"/>
 	</xs:complexType>
+	<xs:complexType name="VehicleBattery">
```
shorowit, line 6398: Seems like most of the elements here are copies of the elements in the `Battery` type, right? Ideally we would have a base `Battery` type with common elements, and then `VehicleBattery` and `HomeBattery` elements would use the base type and add any additional/custom properties. That way we don't accidentally add new elements in the future to vehicles but forget to add them to home batteries. But not critical to change...

schemas/HPXMLBaseElements.xsd
```diff
@@ -6230,6 +6395,56 @@
...
+			<xs:element minOccurs="0" name="GridConnected" type="HPXMLBoolean">
+				<xs:annotation>
+					<xs:documentation>Has the ability to feed electricity back on to the grid.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
+			<xs:element minOccurs="0" name="BuildingConnected" type="HPXMLBoolean">
+				<xs:annotation>
+					<xs:documentation>Has the ability to feed electricity to the building.</xs:documentation>
+				</xs:annotation>
+			</xs:element>
```
shorowit, line 6443: Are these really separate capabilities? An EV could have the ability to be grid connected but not building connected, or vice versa? I guess I assumed it was either connected, in which case it could do both, or neither.

schemas/HPXMLBaseElements.xsd
```diff
@@ -2248,6 +2248,170 @@
...
+															<xs:element minOccurs="0" name="FuelType" type="VehicleFuelType">
+																<xs:annotation>
+																	<xs:documentation>Fuel type of the combustion engine</xs:documentation>
+																</xs:annotation>
+															</xs:element>
+															<xs:element ref="extension" minOccurs="0"/>
+														</xs:sequence>
+													</xs:complexType>
+												</xs:element>
+												<xs:element minOccurs="0" name="HEV">
```
shorowit, line 2342: Probably not necessary, but we could consider adding the `Battery` element here too since these vehicles also have a (small) battery.

schemas/HPXMLBaseElements.xsd
```diff
@@ -2248,6 +2248,170 @@
...
+											<xs:documentation>Number of similar vehicles.</xs:documentation>
+										</xs:annotation>
+									</xs:element>
+									<xs:element minOccurs="0" name="VehicleType">
+										<xs:complexType>
+											<xs:choice>
+												<xs:element minOccurs="0" name="BEV">
+													<xs:annotation>
+														<xs:documentation>Battery electric vehicle</xs:documentation>
+													</xs:annotation>
```
shorowit, line 2284: I personally would have preferred to see the element spelled out and not an acronym. And then you don't need the annotation. @nmerket, any opinion?

schemas/HPXMLBaseElements.xsd
```diff
@@ -2248,6 +2248,170 @@
...
+														<xs:documentation>Internal combustion engine</xs:documentation>
+													</xs:annotation>
+													<xs:complexType>
+														<xs:sequence>
+															<xs:element minOccurs="0" name="FuelType" type="VehicleFuelType"/>
+															<xs:element ref="extension" minOccurs="0"/>
+														</xs:sequence>
+													</xs:complexType>
+												</xs:element>
+												<xs:element minOccurs="0" name="Other">
```
shorowit, line 2364: For completeness, we should probably add fuel cell vehicles above this, since they are available and on the road. And make sure that hydrogen is a possible fuel type.

schemas/HPXMLBaseElements.xsd
shorowit: Also, I think these properties should be moved up from the `Battery` element to the `BEV` element. They are more a property of the car than the battery itself, right?

shorowit (commented):
Nice work, @aspeake. Overall looks pretty good, but I did find some areas that should be improved.

schemas/HPXMLBaseElements.xsd
aspeake: Good question, I think from a technology standpoint they may not be separate, but there may be instances where a home has one and not the other, for example the utility does not offer a program for V2G.

I actually think bidirectional charging is dependent on both the vehicle and the charger, and you would need both to be compatible. Does it make sense to have similar `BuildingConnected` and `GridConnected` attributes for the BEV and the charger?

schemas/HPXMLBaseElements.xsd
aspeake: I think it should be allowed. We do need to allow for 0 in ResStock, as we scale the mileage based on % charged at home, and that can be 0.

schemas/HPXMLBaseElements.xsd
aspeake: Would that result in breaking changes though? Where the home battery elements are now split between the common `Battery` element and the home battery level?

schemas/HPXMLBaseElements.xsd
aspeake: Added in https://github.com/hpxmlwg/hpxml/pull/403/commits/ba511dbb8f4b934c4d9fa06507e11e0a3b1ff739

schemas/HPXMLBaseElements.xsd
aspeake: Added in https://github.com/hpxmlwg/hpxml/pull/403/commits/ba511dbb8f4b934c4d9fa06507e11e0a3b1ff739

schemas/HPXMLBaseElements.xsd
aspeake: I went ahead and spelled them out in https://github.com/hpxmlwg/hpxml/pull/403/commits/ba511dbb8f4b934c4d9fa06507e11e0a3b1ff739

schemas/HPXMLBaseElements.xsd
shorowit: Are we talking about the same thing? The mileage in the car is how many miles it has been driven to date. I don't see why you would use that in ResStock, and I think it's basically impossible to be zero even for a new car.

schemas/HPXMLBaseElements.xsd
shorowit: It wouldn't be breaking, no. It changes the underlying infrastructure but would not change the home battery elements that are exposed to the user.

schemas/HPXMLBaseElements.xsd
aspeake: Ah no, we are not. I was thinking of annual miles, I'll update.

schemas/HPXMLBaseElements.xsd
shorowit: I just realized that it could change the order of elements for the `HomeBattery` though, which would be a breaking change. If that's true, then I suggest not doing this. Nobody likes a breaking change, we should only do it if absolutely necessary. Sorry for the hassle.

schemas/HPXMLBaseElements.xsd
shorowit: It sounds like we need to be clear as to whether these fields describe capabilities or uses. E.g., a vehicle may have V2G capabilities, but it might not be used for a given home.

aspeake: What about adding a new type that looks like:

```
<xs:element minOccurs="0" name="BidirectionalCharging">
	<xs:complexType>
		<xs:sequence>
			<xs:element minOccurs="0" name="BidirectionalChargingEnabled" type="HPXMLBoolean"/>
			<xs:element minOccurs="0" name="GridConnected" type="HPXMLBoolean"/>
			<xs:element minOccurs="0" name="BuildingConnected" type="HPXMLBoolean"/>
		</xs:sequence>
	</xs:complexType>
</xs:element>
```

And apply it both the `ElectricVehicleCharger` and `BatteryElectricVehicle`. This could allow bidirectional charging to be generally enabled but not necessarily connected to a building/grid.

aspeake:
Recent updates include:
- New `FuelCellElectricVehicle` vehicle type
- Optional `MaximumChargingPower` element for the VehicleBattery type
- Include vehicle weight rating (light duty, medium duty, heavy duty)
- Write out vehicle classes instead of using numbers (e.g., passenger car, bus)
- Proposed `BidirectionalCharging` elements for BEV and its charger:
  -  `BidirectionalChargingEnabled`: charger/vehicle has the available technology
  - `GridConnected`: is connected to the grid
  - `BuildingConnected`: is connected to the building

aspeake:
Recent updates include:
- New `FuelCellElectricVehicle` vehicle type
- Optional `MaximumChargingPower` element for the VehicleBattery type
- Include vehicle weight rating (light duty, medium duty, heavy duty)
- Write out vehicle classes instead of using numbers (e.g., passenger car, bus)
- Proposed `BidirectionalCharging` elements for BEV and its charger:
  -  `BidirectionalChargingEnabled`: charger/vehicle has the available technology
  - `GridConnected`: is connected to the grid
  - `BuildingConnected`: is connected to the building

aspeake:
@nmerket thanks in advance for your review, I'd like to request specific feedback on our approach to describe bidirectional charging [here](https://github.com/hpxmlwg/hpxml/pull/403/commits/ba4281baca26ec525dfd45a9e36b33729678ce42). Let me know if you have any suggestions.

schemas/HPXMLBaseElements.xsd
nmerket: I have a lot of the same questions as @shorowit about whether and how this should be included on the car or the EVSE or what. I feel like we're missing a lot of detail on this nascent technology. Is there a use case we're trying to support with this? If so, let's let that inform what it should say. If not, I say we leave it off for now until we have more information.

nmerket (approved):
Looks good to me.

shorowit (approved):

shorowit:
I just updated the images in the PR description.

PR #406, opened 2024-04-16
Adds `UnitHeightAboveGrade` element

shorowit:
Adds `BuildingSummary/BuildingConstruction/UnitHeightAboveGrade` element.

Definition: _"[ft] Height of the unit's lowest conditioned floor above grade. Useful to characterize the height of apartment units above ground level or homes on pier and beam foundations. Can be negative for, e.g., a garden-level apartment unit."_


shorowit:
Need to allow negative values for e.g. garden level apartment units. Clarify that it's the lowest conditioned floor. Change name to `UnitHeightAboveGround`? "Distance" instead? Maybe add a "story" input in the future (e.g., unit on third story).

PR #411, opened 2024-04-25
Adds a `Slab/ExteriorHorizontalInsulation` element

shorowit:
See [here](https://framebuildingnews.com/below-grade-insulation-part-1/) and [here](https://www.renovation-headquarters.com/foundations-29.html) for examples. Sometimes also referred to as "wing" or "skirt" insulation.

Very similar to the existing `PerimeterInsulation` and `UnderSlabInsulation` elements.

image


shorowit:
Add some kind of input for insulation distance below ground. Change name to `ExteriorHorizontalInsulation` so as not to be confused with MH belly/wing foundation or skirts.

PR #413, opened 2024-05-09
Allow negative Elevation values

### LINKED ISSUE #412: Allow negative values for Elevation
It only allows values >= 0 right now.

shorowit:
Fixes #412. Allows `Elevation` field to be negative for locations that are below sea level. The initial implementation incorrectly only allowed values >= 0.


nmerket (approved):

afontani:
I am a little late here, but what about a weather station located in Death Valley, CA. Elevation -282.2′. I think the station in Imperial County California is somewhere near the Salton Sea, which is 235 feet below sea level (elevation -235 ft).

PR #414, opened 2024-06-03
Require > 0 for some R-value elements

shorowit:
Updates `Door/RValue` and `Duct/DuctEffectiveRValue` to require a value greater than zero.


examples/upgrade.xml
```diff
@@ -169,12 +169,12 @@
           <Door>
             <SystemIdentifier id="door1"/>
             <StormDoor>false</StormDoor>
-            <RValue>0.0</RValue>
+            <RValue>2.0</RValue>
```
shorowit, line 172: Why were these set at R-0? Weird...

nmerket (approved):

PR #420, opened 2024-07-11
Electric Panels

### LINKED ISSUE #396: Electric panels
Important when discussing new equipment that potentially requires higher voltage service.

Some initial thoughts from @nmerket:

- Manufacturer
- Model
- Max current rating
- Some way to indicate sub panels
- Whether/how much space is there for new breakers in the panel

Some additional thoughts from Kelly at Radiant Labs:

- Location
- Type (main/sub panel)
- Count of total breakers
- Count of breakers used 
- ID (plus idrefs to relate other systems to the panel -  not sure if you'd want to get this complicated)
- Voltage (some panels in older homes can only support 120V)

nmerket:
Fixes #396.

Needed for HOMES rebates. Keeping it very simple for now. There's a lot of good ideas below. I created a new issue #422 to add those in later.

HPXML_ElectricPanels

### DIFF
--- a/merged_schema/HPXMLMerged.xsd
+++ b/merged_schema/HPXMLMerged.xsd
@@ -6296,18 +6296,6 @@
 					<xs:documentation>[A]</xs:documentation>
 				</xs:annotation>
 			</xs:element>
-			<xs:element maxOccurs="unbounded" minOccurs="0" name="BranchCircuit">
-				<xs:complexType>
-					<xs:sequence>
-						<xs:element minOccurs="0" name="BreakerAmperage"/>
-						<xs:element maxOccurs="1" minOccurs="0" name="SubPanel" type="ElectricPanelType">
-							<xs:annotation>
-								<xs:documentation>Description of a sub-panel.</xs:documentation>
-							</xs:annotation>
-						</xs:element>
-					</xs:sequence>
-				</xs:complexType>
-			</xs:element>
 			<xs:element minOccurs="0" ref="extension"/>
 		</xs:sequence>
 		<xs:attribute name="dataSource" type="DataSource"/>
--- a/merged_schema/HPXMLMerged.xsd
+++ b/merged_schema/HPXMLMerged.xsd
@@ -2147,34 +2147,7 @@
 			<xs:element minOccurs="0" name="ElectricPanels">
 				<xs:complexType>
 					<xs:sequence>
-						<xs:element maxOccurs="unbounded" name="ElectricPanel">
-							<xs:complexType>
-								<xs:sequence>
-									<xs:group ref="SystemInfo"/>
-									<xs:element minOccurs="0" name="Manufacturer" type="Manufacturer"/>
-									<xs:element minOccurs="0" name="Model" type="Model"/>
-									<xs:element minOccurs="0" name="Location" type="BatteryLocation"/>
-									<xs:element minOccurs="0" name="Voltage" type="ElectricPanelVoltage"/>
-									<xs:element minOccurs="0" name="MaxCurrentRating" type="Current">
-										<xs:annotation>
-											<xs:documentation>[A]</xs:documentation>
-										</xs:annotation>
-									</xs:element>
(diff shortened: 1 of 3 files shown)


nmerket:
We discussed having open breaker slots or some sense of breakers. I haven't tackled that yet.

Another option for sub panels is that I could make the panel a type and have it be a hierarchy instead of a reference. That would enforce the relationship a little more explicitly.

nmerket:
- How can we represent smart panels in this structure? Noel to reach out to NREL team working on smart panels.

nmerket:
Recommendation from @chrisbalbach:

Look at [how Revit represents these systems](https://help.autodesk.com/view/RVT/2025/ENU/?guid=GUID-89BE904A-E0EF-4BF7-B7FC-1E504584637C) and pattern this after that.

lixiliu:
Is the idea to make the panel an object with attributes only or an object that can do accounting of equipment and area-based loads and breaker slot requirement?

Another consideration is multi-family structure. For whole-building simulation, we may want the unit's panel to be able to connect to the building's main panel. Is the schematic expandable like that?

I am not familiar with the HPXML schematics, hopefully they are relevant questions.

jmaguire1:
This is actually super helpful to support some work Lixi is scoping out for next year!

+1 on being able to have breaker slots. You can be limited by either not having the amperage or the # of breaker slots in terms of trying to electrify, so we want to be able to have both represented.

Similar to what Lixi said: do we want to report not just the panel amperage, but how much is currently in use? Ditto on breaker slots. That's what we'd want for figuring out how constrained you are when considering electrification. We could do that calculation based on the equipment in the home and NEC (electric code), but I think we'd want to store that in this element?

We could use manufacturer or model to work out if it's smart or not, rather than having a separate element for smart panels. I'm not opposed to that approach and it's one less element to worry about. I'm also not sure exactly what we'd change in OS-HPXML if you did select a smart panel? I think ideally that'd allow some loads to be shifted around, but that'd have to happen when schedules are generated I think? @rajeee probably has better thoughts about what we'd actually do with that information if we had it in E+.

This is maybe going too far, but: there are a lot of options like circuit splitters (for example [this one](https://getneocharge.com/products/neocharge-smart-splitter)) that people are looking at as ways to avoid a panel upgrade. The basic idea is it'd let you share a circuit between two devices that aren't likely to run at the same time, like say a dryer and EV, or multiple EVs. Do we want to be able to represent that here? Maybe could/should be an extension element if we need it?

rajeee:
Happy to contribute.

The features of Smart panel that distinguishes it from regular panels are:

1. Internet/App connected.
2. Circuit level monitoring and manual control 
3. Automatic circuit control features (such as turning off EV charger circuit if the total load exceed a certain threshold), or scheduled circuit control
4. Advanced energy management during outage (When Battery and Solar is connected, it can coordinate which circuits to power up based on battery SOC)
5. Automatic "fault" detection (for example alert you of large and unexpected increase in load in a particular circuit - perhaps forgot to turn off the space heater)

Most of these functionalities are more in the modelling side rather than in schema side. However, I feel like it might be worth definining a separate circuit element that is a child to the panel element.

The circuit element would contain:
ID: The circuit id
slot_number: The slot number of the panel it is in
parent: The panel
breaker_capacity: The amperage of the breaker used for this circuit
smart_control: True/False (If we want to model this feature)


And I am imagining, all the different loads in the house could point to a particular circuit id that they are connected to. This would allow us to not only calculate the panel total electricity, but also aggregate the consumption at circuit level. But, actually figuring this out (how many circuits and what breaker capacity for each, and which load is connected to what circuit) is definitely a challenge. But at least the schema would support that if someone wanted to do a detailed modelling like that.

lixiliu:
> Fixes #396.
> 
> Needed for HOMES rebates.
> 
> HPXMLBaseElements_ElectricPanel

@nmerket - A subpanel is wired to the main via a double-pole branch circuit, so this is consistent with your schematic. I would add:

- to the branch circuit a parameter for voltage drop (1-pole vs. 2-pole),
- to the electrical panel a parameter for the number of breaker slots.


Thanks for taking my feedback!

PR #421, opened 2024-07-15
Additional water heater elements

shorowit:
Adds `SealedCombustion`, `CondensingSystem`, `AtmosphericBurner`, and `PowerBurner` elements for water heaters; these elements already exist for furnaces/boilers.

### DIFF
--- a/schemas/HPXMLBaseElements.xsd
+++ b/schemas/HPXMLBaseElements.xsd
@@ -1802,6 +1802,9 @@
 									<xs:element minOccurs="0" name="HasSharedCombustionVentilation" type="HPXMLBoolean"/>
 									<xs:element minOccurs="0" name="CombustionVentilationOrphaned" type="HPXMLBoolean"/>
 									<xs:element name="CombustionVentingSystem" minOccurs="0" type="LocalReference"/>
+									<xs:element minOccurs="0" name="CondensingSystem" type="HPXMLBoolean"/>
+									<xs:element minOccurs="0" name="AtmosphericBurner" type="HPXMLBoolean"/>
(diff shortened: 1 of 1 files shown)


shorowit:
@jmaguire1 Can you look these over and provide feedback? I'm not sure if we should use, e.g., "DirectVent" or "PowerVent" instead.

jmaguire1:
Yes, I never heard from anyone at GTI on this. It is a little confusing, but I think the way we'll want to do this:

CondensingSystem: If a tank WH, TRUE if UEF > 0.77 (or corresponding EF). If tankless, TRUE if EF (or UEF) > 0.94. Else False. 

AtmosphericBurner: If tank WH, UEF < 0.64. FALSE for any tankless.
PowerBurner: The opposite of AtmosphericBurner, so TRUE if UEF >= 0.64 and for any tankless.
SealedCombustion: Same as "PowerBurner"


to rewrite the code for skinlossfrac [here](https://github.com/NREL/OpenStudio-HPXML/blob/c5fa0c65c969dcce6aa6b19cfa42c2a2fa93c015/HPXMLtoOpenStudio/resources/waterheater.rb#L2024), I think this makes it:

if "AtmosphericBurner":
    skinlossfrac = 0.63
if "Condensing":
    skinlossfrac = 0.96
else:
    skinlossfrac = 0.91.

What really matters here for whether you fall in the 0.91 category is if there's a flue damper, which I think is true pretty much any time "AtmosphericBurner" is false. Condensing vent a little colder, which is how we ended up with a slightly higher skinlossfrac here.

shorowit:
A good resource from AO Smith: https://www.hotwater.com/info-center/water-heater-venting.html

PR #428, opened 2024-10-16
Measure Taxonomy

nmerket:
Stubbing out a new optional `<MeasureType>` element that can add more formalized description of complicated measures. MeasureType is a choice, meaning only one of the sub elements can be selected. Most measures would use the `<InstallReplace>` type, which means what it does now.

HPXMLBaseElements_MeasureType

An example of how this might look in a measure is:

```xml
                <Measure>
                    <MeasureSystemIdentifiers>
                        <SystemIdentifiersInfo id="measure1"/>
                    </MeasureSystemIdentifiers>
                    <MeasureType>
                        <AtticEncapsulation>
                            <DuctsMovedToConditionedSpace>true</DuctsMovedToConditionedSpace>
                        </AtticEncapsulation>
                    </MeasureType>
                    <ReplacedComponents>
                        <ReplacedComponent idref="ducts-before"/>
                        <!-- This wouldn't actually mean the ducts were replaced, but that they were modified in the measure -->
                    </ReplacedComponents>
                    <InstalledComponents>
                        <InstalledComponent idref="ducts-after"/>
                        <!-- The change to ducts would likely only be the Location element -->
                    </InstalledComponents>
                </Measure>
```

@GamalielL


shorowit:
I was originally thinking this could replace all the [Improvement](https://hpxml.nrel.gov/search/?query=improvement&version=101&expanded=False) elements (e.g., HVACDistributionImprovement). However, I realize now that there are two related but slightly different use cases. One is that you might have a description of a building at the present and want to know if, e.g., the ducts have ever been sealed in the past. The other is that you are describing pre/post buildings or proposed upgrades where the duct sealing measure has been applied.

Is it possible/appropriate to use the `Measure` element to describe the former use case?

GamalielL:
That's an interesting point, Scott. There's obviously some need for fields that denote past improvements, as with the HOMES requirement to indicate whether a home was previously professionally air sealed.  I don't think the Measure element is the appropriate method for indicating past improvements.  Most of what is in those Improvement items has either been made obsolete by other elements or should be handled in measures, but some of it could be useful for clarifying the history of a home. For instance, it's useful to know whether duct or air sealing have previously been performed in an building where safety concerns prevent testing.

nmerket:
Make `DuctSealing` and `DuctInsulation` boolean.

merged_schema/HPXMLMerged.xsd
```diff
@@ -4894,6 +4894,69 @@
...
+							<xs:complexType>
+								<xs:sequence>
+									<xs:element minOccurs="0" name="DuctInsulation" type="HPXMLBoolean">
+										<xs:annotation>
+											<xs:documentation>Add ids to the duct insulation (breaking change)?</xs:documentation>
+										</xs:annotation>
+									</xs:element>
+									<xs:element minOccurs="0" name="DuctSealing" type="HPXMLBoolean">
+										<xs:annotation>
+											<xs:documentation>Add ids to the duct sealing or leakage (breaking change)?</xs:documentation>
```
shorowit, line 4950: @nmerket It looks like this is a reminder to consider adding IDs (a breaking change). Should we just delete these comments for now and consider this change for a future HPXML version?

shorowit (commented):
I just updated the image in the PR. The two checklist items are clearly addressed. Just one issue below.

merged_schema/HPXMLMerged.xsd
nmerket: Do you remember the context for this @GamalielL?

GamalielL: This was about adding IDs to the DuctInsulation and DuctLeakage sections so that they could be referenced as installed/replaced components.  This would make the DuctImprovement choice unnecessary, because duct insulation and sealing measures could be distinguished by the referenced components.

schemas/HPXMLBaseElements.xsd
```diff
@@ -4880,6 +4880,69 @@
...
+									<xs:element minOccurs="0" ref="extension"/>
+								</xs:sequence>
+							</xs:complexType>
+						</xs:element>
+						<xs:element name="HVACRedesign">
+							<xs:complexType>
+								<xs:sequence>
+									<xs:element minOccurs="0" name="DistributionTypeChange" type="HPXMLBoolean"/>
+									<xs:element minOccurs="0" name="PartialDisplacement" type="HPXMLBoolean"/>
+									<xs:element minOccurs="0" name="HotWaterIntegrated">
```
GamalielL, line 4914: We should add documentation of these elements. I've proposed documentation below, but I'm not sure it covers all possibilities.

- DistributionTypeChange=TRUE indicates that an HVAC replacement necessitates a change to the distribution type
- PartialDisplacement=TRUE indicates that the measure references and installed component that partially displaces the load met by an existing component that will remain installed, but meet a smaller load
- HotWaterIntegrated=TRUE indicates that the installed system also supplies domestic hot water

merged_schema/HPXMLMerged.xsd
nmerket: Should we make that change or leave it as is with the DuctImprovement choice?

shorowit: Can we make a decision here so we can move forward? I vote for leave as is (by which I mean removing the `xs:documentation` comments), mostly because I think we want to avoid a breaking change. But I don't have a strong opinion.

nmerket:
@shorowit @GamalielL I think this is ready to go. Let me know if you think otherwise.

merged_schema/HPXMLMerged.xsd
```diff
@@ -4895,6 +4895,72 @@
...
+							<xs:complexType>
+								<xs:sequence>
+									<xs:element ref="extension" minOccurs="0"/>
+								</xs:sequence>
+							</xs:complexType>
+						</xs:element>
+						<xs:element name="AtticEncapsulation">
+							<xs:complexType>
+								<xs:sequence>
+									<xs:element minOccurs="0" name="DuctsMovedToConditionedSpace" type="HPXMLBoolean"/>
```
GamalielL, line 4911: 'true' indicates that the attic encapsulation measure brought existing ducts inside the conditioned envelope.

merged_schema/HPXMLMerged.xsd
```diff
@@ -4895,6 +4895,72 @@
...
+									<xs:element minOccurs="0" ref="extension"/>
+								</xs:sequence>
```
GamalielL, line 4913: AtticEncapsulation measures move the thermal boundary of the building from the floor of an attic space to the roof and gable walls, converting the attic from unconditioned to conditioned.

merged_schema/HPXMLMerged.xsd
```diff
@@ -4895,6 +4895,72 @@
...
+							</xs:complexType>
+						</xs:element>
+						<xs:element name="FoundationEncapsulation">
+							<xs:complexType>
+								<xs:sequence>
+									<xs:element minOccurs="0" name="DuctsMovedToConditionedSpace" type="HPXMLBoolean"/>
+									<xs:element minOccurs="0" ref="extension"/>
+								</xs:sequence>
```
GamalielL, line 4921: FoundationEncapsulation measures move the thermal boundary of the building from the floor above the foundation space to the slab and foundation walls, converting the foundation from unconditioned to conditioned.

PR #431, opened 2024-12-10
HPWH additional fields

### LINKED ISSUE #404: HPWH additional fields
Some way to describe:
- 120V products (could apply to other appliances like dryers, cooking range, etc. too)
- Shared vs dedicated circuit
- Tempering valve (integrated vs separate, temperature, etc.)
- ???

shorowit:
Closes #404. Adds elements to `WaterHeatingSystem`:

### Voltage

`HPWHVoltage`: Enumeration with choices of: "240V", "120V", "120V, dedicated circuit", "120V, shared circuit", and "other". 

image  

### Mixing valve setpoint

`MixingValueSetpoint`: Temperature > 0.

image

### DIFF
--- a/schemas/HPXMLBaseElements.xsd
+++ b/schemas/HPXMLBaseElements.xsd
@@ -1737,6 +1737,7 @@
 										</xs:annotation>
 									</xs:element>
 									<xs:element minOccurs="0" name="HPWHOperatingMode" type="HPWHOperatingMode"/>
+									<xs:element minOccurs="0" name="HPWHVoltage" type="HPWHVoltage"/>
 									<xs:element minOccurs="0" name="HPWHDucting">
 										<xs:complexType>
 											<xs:sequence>
@@ -1831,6 +1832,7 @@
 										</xs:annotation>
 									</xs:element>
 									<xs:element minOccurs="0" name="HasMixingValve" type="HPXMLBoolean"/>
+									<xs:element minOccurs="0" name="MixingValueSetpoint" type="TemperatureGreaterThanZero"/>
 									<xs:element minOccurs="0" name="UsesDesuperheater" type="HPXMLBoolean">
 										<xs:annotation>
 											<xs:documentation>Indicates whether this water heater uses a desuperheater. The attached heat pump or air conditioner can be referenced in the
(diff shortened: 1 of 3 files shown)


jmaguire1:
This covers everything I was looking for with 120V HPWHs (and the mixing valve setpoint has some bonus implications for covering some edge cases with load shifting controls too). Thanks @shorowit for implementing this! We'll talk about it with the working group in a few days and make sure they're on board, but definitely meets all my needs.

schemas/HPXMLBaseElements.xsd
```diff
@@ -1831,6 +1832,7 @@
 										</xs:annotation>
 									</xs:element>
 									<xs:element minOccurs="0" name="HasMixingValve" type="HPXMLBoolean"/>
+									<xs:element minOccurs="0" name="MixingValueSetpoint" type="TemperatureGreaterThanZero"/>
```
nmerket, line 1835: What about changing this to

```suggestion
									<xs:element minOccurs="0" name="MixingValueOutletTemperature" type="TemperatureGreaterThanZero"/>
```

to be consistent with `HotWaterTemperature`?

schemas/HPXMLDataTypes.xsd
```diff
@@ -5053,6 +5053,22 @@
 			</xs:extension>
 		</xs:simpleContent>
 	</xs:complexType>
+	<xs:simpleType name="HPWHVoltage_simple">
+		<xs:restriction base="xs:string">
+			<xs:pattern value="240V"/>
+			<xs:pattern value="120V"/>
+			<xs:pattern value="120V, dedicated circuit"/>
+			<xs:pattern value="120V, shared circuit"/>
```
nmerket, line 5061: Nit picky, but I like it without the extraneous commas.

```suggestion
			<xs:pattern value="120V dedicated circuit"/>
			<xs:pattern value="120V shared circuit"/>
```

schemas/HPXMLBaseElements.xsd
jmaguire1: Maybe nitpicky, but: it won't always maintain the setpoint temperature in practice. If for whatever reason the tank outlet temp is colder than the mixing valve setpoint temp, it won't really be the outlet temperature. I like setpoint slightly better for that reason, but it is definitely nitpicky.

shorowit: We'll discuss this w/ Noel when he gets back and then this should be good to merge.

PR #435, opened 2025-01-09
ElectricPanel: add BranchCircuits and ServiceFeeders

joseph-robertson:
Add 4 new fields to ElectricPanel: `HeadroomSpaces`, `RatedTotalSpaces`, `BranchCircuits`, and `ServiceFeeders`.

ElectricPanel:
* `HeadroomSpaces` (integer greater than or equal to zero)
* `RatedTotalSpaces` (integer greater than zero)

image

`BranchCircuits` mirrors a main panel (or subpanel) breaker box:
  * voltage (120, 240)
  * max current rating (A)
  * occupied spaces (0.0, 0.5, 1.0, 2.0)
  * attached component(s)
  * attached subpanel

image

`ServiceFeeders` supports service load calculations (e.g., 2023 NEC load-based):
  * load type (heating, cooling, hot water, clothes dryer, dishwasher, range/oven, refrigeration, mech vent, permanent spa heater, permanent spa pump, pool heater, pool pump, well pump, electric vehicle charging, lighting, kitchen, laundry, other)
  * power rating (W, double greater than or equal to zero)
  * new load (boolean)
  * number of appliances (integer greater than or equal to zero)
  * attached component(s)

image


shorowit (commented):
Haven't looked very closely, just happened to notice these.

schemas/HPXMLBaseElements.xsd
```diff
@@ -6611,8 +6611,84 @@
...
+		</xs:sequence>
+		<xs:attribute name="dataSource" type="DataSource"/>
+	</xs:complexType>
+	<xs:complexType name="ServiceFeederType">
+		<xs:sequence>
+			<xs:group ref="SystemInfo"/>
+			<xs:element minOccurs="0" name="LoadType" type="ServiceFeederLoadType"/>
+			<xs:element minOccurs="0" name="PowerRating" type="HPXMLDouble">
+				<xs:annotation>
+					<xs:documentation>Volt-Amps [VA] or Watts [W]</xs:documentation>
```
nmerket, line 6681: We should just specify one unit here. There's [a difference between VA and W](https://www.eaton.com/de/en-gb/products/backup-power-ups-surge-it-power-distribution/backup-power-ups/va-versus-watts--eaton.html). Which is most important for this use case?

nmerket (commented):
Looks pretty good. One comment about units below. Also, does it make sense to have some connection between `BranchCircuits` and `ServiceFeeders`?

schemas/HPXMLBaseElements.xsd
joseph-robertson: Just a thought... support both? Similar to, e.g., AnnualHeatingEfficiency for HeatingSystem:

```
<AnnualHeatingEfficiency>
  <Units>AFUE</Units>
  <Value>0.92</Value>
```
```
<PowerRating>
  <Units>VA or W</Units>
  <Value>0.92</Value>
```

shorowit: If you're going to allow both in the schema, then you would need to update OS-HPXML to handle both. It seems like you currently assume it's W? https://github.com/NREL/OpenStudio-HPXML/blob/electric_panel/docs/source/workflow_inputs.rst?plain=1#L4700

joseph-robertson: I would? Couldn't I just exclude VA using the EPvalidator?

lixiliu: For load calculation, it's both and but it's used interchangeably there and do not distinguish between real and reactive portion of the power. If we can't support both, then just go with W.

nmerket: Let's just go with W if for the most common use case they're not differentiating.

lixiliu:
> Looks pretty good. One comment about units below. Also, does it make sense to have some connection between `BranchCircuits` and `ServiceFeeders`?

`BranchCircuits` and `ServiceFeeders` are loosely connected by the systems they share but they should not be connected directly as we don't always count all the loads on a single branch circuit (e.g., if a branch circuit has a dishwasher and a gas water heater, the service feeder would only want to take the dishwasher load).

schemas/HPXMLBaseElements.xsd
joseph-robertson: Done in [d347083](https://github.com/hpxmlwg/hpxml/pull/435/commits/d34708353c555ab33015296db5c59dc8a6fc182b).

schemas/HPXMLBaseElements.xsd
```diff
@@ -6611,8 +6611,84 @@
...
+				</xs:annotation>
+			</xs:element>
+		</xs:sequence>
+		<xs:attribute name="dataSource" type="DataSource"/>
+	</xs:complexType>
+	<xs:complexType name="ServiceFeederType">
+		<xs:sequence>
+			<xs:group ref="SystemInfo"/>
+			<xs:element minOccurs="0" name="LoadType" type="ServiceFeederLoadType"/>
+			<xs:element minOccurs="0" name="PowerRating" type="HPXMLDouble">
```
shorowit, line 6679: Should this be required to be greater than zero?

schemas/HPXMLDataTypes.xsd
```diff
@@ -5269,6 +5268,76 @@
...
+			<xs:enumeration value="branch circuit"/>
+		</xs:restriction>
+	</xs:simpleType>
+	<xs:complexType name="DemandLoadsDemandLoadType">
+		<xs:simpleContent>
+			<xs:extension base="DemandLoadsDemandLoadType_simple">
+				<xs:attribute name="dataSource" type="DataSource"/>
+			</xs:extension>
+		</xs:simpleContent>
+	</xs:complexType>
```
shorowit, line 5340: It looks like these are unused?

schemas/HPXMLBaseElements.xsd
joseph-robertson: https://github.com/hpxmlwg/hpxml/pull/435/commits/bffc5936e7975ef57189d78b38b04e1925d8032d

joseph-robertson:
> Is it possible to rename `Headroom` as `HeadroomSpaces` in the schema, this way it's clear that it's about space? We use the term headroom in both capacity and space calculations.

https://github.com/hpxmlwg/hpxml/pull/435/commits/bffc5936e7975ef57189d78b38b04e1925d8032d

schemas/HPXMLBaseElements.xsd
joseph-robertson: I think this should be greater than or equal to zero.

PR #447, opened 2025-06-02
Fix mixing valve typo

shorowit:
Sorry @nmerket, I was looking over the 4.2-rc1 changes and just noticed this typo. Apparently this slipped past through sets of eyeballs in #431...


nmerket (approved):
Good catch.

PR #471, opened 2026-03-25
Skylight curbs/shafts

shorowit:
Allows describing details of curbs/shafts under `Skylight`:
- Adds `Curb` element with properties
- Adds `AttachedToKneeWall` element
- Adds `CeilingAreaDisplaced` element

image

Example curb:
image

Example shaft:
3dda79e7c0f105f87e0ca5169d05104d-710899820


merged_schema/HPXMLMerged.xsd
```diff
@@ -8748,6 +8800,38 @@
...
+		<xs:simpleContent>
+			<xs:extension base="SkylightCurbMaterial_simple">
+				<xs:attribute name="dataSource" type="DataSource"/>
+			</xs:extension>
+		</xs:simpleContent>
+	</xs:complexType>
+	<xs:simpleType name="SkylightShaftMaterial_simple">
+		<xs:restriction base="xs:string">
+			<xs:enumeration value="metal"/>
+			<xs:enumeration value="plywood"/>
```
shorowit: Add gypsum or drywall (see what's used elsewhere).

merged_schema/HPXMLMerged.xsd
```diff
@@ -1395,6 +1395,58 @@
...
+												<xs:element minOccurs="0" name="AssemblyEffectiveRValue" type="RValueGreaterThanZero">
+													<xs:annotation>
+														<xs:documentation>This should indicate the effective R-value of the complete assembly including any air films or other treatments. For below-grade surfaces adjacent to ground, it should not include the insulating effect of the ground.</xs:documentation>
+													</xs:annotation>
+												</xs:element>
+												<xs:element ref="extension" minOccurs="0"/>
+											</xs:sequence>
+										</xs:complexType>
+									</xs:element>
+									<xs:element minOccurs="0" name="Shaft">
```
shorowit: Consider how a Shaft is different from a KneeWall (if at all).

shorowit (commented):
Comments below.

shorowit (commented):
This is probably also the right time to address the fact that the skylight penetration area through the roof can be different than the ceiling (floor) penetration area, in the case where a shaft widens as it descends.

RockyMtnBrian:
On the matter of skylight shafts:  I think they can readily be handled as kneewalls (walls between attic and living space).  The only case I can think of that may seem odd would be suntube-style products, where the light shaft is essentially a metal tube that is round in cross section.  

As for the skylight penetration of the ceiling (i.e. attic floor), I suggest an element under Skylight, affiliated with the AttachedToFloor element.  To collect the ceiling area (attic floor) that is displaced by the skylight penetration, perhaps the label CeilingAreaDisplaced would work?

Using a default value of Skylight Area for CeilingAreaDisplaced could work, since it generally will overstate the latter.  Since the attic insulation is generally aligned with the ceiling surface, there is built-in incentive to not rely on the default.

shorowit:
One more idea: Add an `AttachedToWall` so that you can relate a kneewall with a given skylight (shaft)?

RockyMtnBrian:
I was trying to figure out why you wanted an AttachedToWall;  is it because Reference Home configuration requires removing skylights, and you want to clear out the associated skylight shaft as well as restore the opening in the ceiling?

shorowit:
No, I don't think it would be used for an ERI calculation. My reasoning is that I originally added skylight shaft inputs to support Manual J because that's how it refers to these surfaces (see example language below), and I _think_ that when Manual J design loads are reported by component type, these heat losses are associated with the skylight. But I'm not 100% sure, I will need to double-check.

image

shorowit:
I updated the proposal based on everyone's feedback; see the updated description at the top. I basically:
1. Replaced `Skylight/Shaft` with a `Skylight/AttachedToKneeWall` element
2. Added a `CeilingAreaDisplaced` element
3. Added/updated some descriptions

shorowit:
Working group feedback:
- [x] Document that `Floor/Area` should be gross area, exclusive of any penetrations like skylight shafts
- [x] Clarify skylight curb area (depth above roof vs rafter height); check Manual J

shorowit:
Changes made:
- Added `Curb/Height` element (description: "[in] Height above roof plane") as an alternative property
- Renamed `Curb/Area` to `Curb/SurfaceArea` to make it clearer that it's not the rough opening area (which is the `Skylight/Area` element) and updated description to "[sq.ft.] Curb height above roof plane multiplied by the perimeter"
- Updated description of `Floor/Area`: "[sq.ft.]" => "[sq.ft.] Gross floor area, e.g., includes skylight shaft or sun tunnel area penetration through ceiling (attic floor)"
- Updated description of `Wall/Area`: "[sq.ft.] Gross wall area" => "[sq.ft.] Gross wall area, e.g., includes window/door area"

PR #472, opened 2026-03-26
Add "outdoor closet" and "outside" choices for `UnitLocation`

shorowit:
Allows a more explicit description for the water heater location. Currently the closest choices are "mechanical closet" (which I assume to be inside) and "other exterior" (which could mean anything).

Note that this HPXML data type is also used for HVAC systems.

**"outdoor closet"** examples:
image
image

**"outside"** examples:
image
image


merged_schema/HPXMLMerged.xsd
```diff
@@ -9407,6 +9407,8 @@
 			<xs:enumeration value="other interior"/>
 			<xs:enumeration value="other multifamily buffer space"/>
 			<xs:enumeration value="other non-freezing space"/>
+			<xs:enumeration value="outdoor closet"/>
```
shorowit, line 9457: See if we can add a definition for this.

shorowit: Don't include "manufactured home" in "outdoor closet" since it can show up in standard SFD homes too.

